PR #12, opened 2017-07-15
Let's get Calico running so that Kubernetes actually starts up.

darron:
Hat tip to https://github.com/kelseyhightower/kubeadm-single-node-cluster


kris-nova:
Thanks @kelseyhightower for the tip!

PR #18, opened 2017-07-15
Fix Makefile ordering and remove bootstrap.go

k4k:
Removes bootstrap.go and flips default ordering


k4k:
@darron Sorry about that. Thought I was adding the removal of the file, not adding it back in. Fixed.

Issue #11, opened 2017-07-15
New user on-boarding


mhausenblas:
It would be great to have a walkthrough supporting new users how to do the basic steps

geojaz:
closed by #20

PR #20, opened 2017-07-15
addresses #11

mhausenblas:
This is the initial 101 walkthrough for AWS


docs/walkthrough.md
```diff
@@ -0,0 +1,87 @@
+# Walkthrough
+
+In the following, we gonna show you how to use `kubicorn` to ramp up a Kubernetes cluster in AWS, use it and tear it down again.
```
k4k: Please fix phrasing "we gonna show" to "we're going to show"

docs/walkthrough.md
```diff
@@ -0,0 +1,87 @@
...
+
+```
+$ cat _state/myfirstk8s/cluster.yaml
+```
+
+We're now in a position to have the cluster resources defined, locally, based on the selected profile. Next we will apply the so defined resources using the `apply` command, but before we do that we'll set up the access to AWS. You might want to create a new [IAM user](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_users_create.html) for this with the following permissions:
+
+![AWS IAM permissions required for kubicorn](img/aws-iam-user-perm-screen-shot.png)
+
+Next, export the two environment variables `ACCESS_KEY_ID` and `SECRET_ACCESS_KEY` so that `kubicorn` can pick it up in the next step:
```
geojaz: I think this should be `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`... does this work?

docs/walkthrough.md
```diff
@@ -0,0 +1,87 @@
...
+The first thing you will do now is to define the cluster resources. For this, you need to select a certain profile. Of course, once you're more familiar with `kubicorn`, you can go ahead and extend existing profiles or create new ones.
+In the following we'll be using an existing profile called `aws`, which is—surprise, surprise—a profile for a cluster in AWS.
+
+Now execute the following command:
+
+```
+$ kubicorn create --name myfirstk8s --profile aws
+```
+
+Note that `kubicorn` executes silently as long as there are no errors but you can always use the `--verbose` flag to increase the information output, for example `kubicorn create --verbose 4` gives you detailed info on what it is doing at any step.
```
k4k: #23 will change the default behavior to include helpful next step.

docs/walkthrough.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+The first thing you will do now is to define the cluster resources. For this, you need to select a certain profile. Of course, once you're more familiar with `kubicorn`, you can go ahead and extend existing profiles or create new ones.
+In the following we'll be using an existing profile called `aws`, which is—surprise, surprise—a profile for a cluster in AWS.
+
+Now execute the following command:
+
+```
+$ kubicorn create --name myfirstk8s --profile aws
+```
```
heavenlyhash: Perhaps add a comment that this doesn't Do Things on the cloud yet, it just generates local files with the intentions which later command really apply.  (I figured it out, but not until scrolling down and noticing "apply" and "oh I didn't give it credentials yet, riiight")

Issue #17, opened 2017-07-15
Commands should provide some context on execution


mhausenblas:
Currently, when running for example `kubicorn create` the execution by default is silent. This is not a good UX, new users might think nothing happened. So after execution, the command should output a summary and suggest next steps, for example:

```
$ kubicorn create --name myfirstk8s --profile aws
I've created _state/myfirstk8s/cluster.yaml based on your input and now you can run kubicorn apply
```

PR #23, opened 2017-07-15
Add user feedback to create

k4k:
`kubicorn create` was previously silent on completion reported in #17
This adds feedback and next steps to run `kubicorn apply`


mhausenblas:
LGTM, let's 🚢 it!

k4k:
Before we merge this I'd like to include the same type of output for `kubicorn apply`

PR #33, opened 2017-07-15
Add list sub-command

k4k:
* Adds a subcommand to list the states present in the state store


mhausenblas:
LGTM! In general, how're we doing w.r.t. test cases? CC: @kris-nova

Issue #13, opened 2017-07-15
Signal handler and support for AWS


geojaz:
and try to die nicely instead of aborting

Issue #10, opened 2017-07-15
Unable to fully delete cluster


geojaz:
```
erichole   $ kubicorn delete -n eric -v 4
2017-07-15T11:14:25-06:00 [✔]  Selected [fs] state store
2017-07-15T11:14:25-06:00 [▶]  asg.Actual
2017-07-15T11:14:26-06:00 [▶]  asg.Delete
2017-07-15T11:14:26-06:00 [✔]  Deleted ASG [eric.amazon-node]
2017-07-15T11:14:26-06:00 [▶]  lc.Actual
2017-07-15T11:14:27-06:00 [✖]  Unable to destroy resources for cluster [eric]: Found [0] Launch Configurations for ID [eric.amazon-node]
```

geojaz:
relates to #13

kris-nova:
Closing as I think this is fixed now - right @geojaz?

PR #39, opened 2017-07-16
Intercepts ctrl-C and unwinds whatever work was happening

geojaz:


geojaz:
Closes #13 and maybe #10 (because we _shouldn't_ get into these situations anymore...)

bootstrap/bootstrap.go
```diff
@@ -1,283 +0,0 @@
-// Code generated by go-bindata.
```
kris-nova: Thanks for removing this!

kris-nova:
lgtm

PR #40, opened 2017-07-16
More targets for gofmt and gofmt all the things

geojaz:


kris-nova:
oh hell yeah - we need `go fmt` in the make file too

PR #50, opened 2017-07-16
DigitalOcean support

kris-nova:
Adding implementation for Ubuntu on DigitalOcean


kris-nova:
Pre-emptive merge to get docs in

Issue #51, opened 2017-07-16
install fails with `undefined: bootstrap.Asset`


paulczar:
When I try to install I get an error about bootstap.Asset being undefined.  I believe its looking for a function inside the bootstrap.go file that was removed here - https://github.com/kris-nova/kubicorn/pull/21

```
$ go get github.com/kris-nova/kubicorn
# github.com/kris-nova/kubicorn/cloud/amazon/resources
../go/src/github.com/kris-nova/kubicorn/cloud/amazon/resources/launchconfiguration.go:165: undefined: bootstrap.Asset
```

kris-nova:
Checking out now

kris-nova:
I will open up another issue for the bootstrap.go concern now that we realize it's breaking `go get`

Thanks for finding this!

PR #53, opened 2017-07-16
Remove some hard coded paths in Makefile

paulczar:
* There were a few hardcoded paths that seemed unecessary as they can be referred to from current directory
* Attempt to deal with go-bindata not being present ( took me a while to figure out what was going on )
* Make `bin/` dir before building with docker so at least its owned by non-root even if binaries are.
* Add bin/ to gitignore
* Update readme with install options

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 _state*
 .idea*
 .DS_Store
+bin/*
 bootstrap/bootstrap.go
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,22 @@ PKGS=$(shell go list ./... | grep -v /vendor)
 SHELL_IMAGE=golang:1.8.3
 GIT_SHA=$(shell git rev-parse --verify HEAD)
 VERSION=$(shell cat VERSION)
+PWD=$(shell pwd)
 
 default: bindata compile
+
+all: default install
+
 compile:
-	@go build -o ${GOPATH}/bin/kubicorn -ldflags "-X github.com/kris-nova/kubicorn/cmd.GitSha=${GIT_SHA} -X github.com/kris-nova/kubicorn/cmd.Version=${VERSION}" main.go
+	go build -o bin/kubicorn -ldflags "-X github.com/kris-nova/kubicorn/cmd.GitSha=${GIT_SHA} -X github.com/kris-nova/kubicorn/cmd.Version=${VERSION}" main.go
+
+install:
+	install --mode=0755 bin/kubicorn ${GOPATH}/bin/kubicorn
 
 bindata:
+	which go-bindata > /dev/null || go get -u github.com/jteeuwen/go-bindata/...
 	rm -rf bootstrap/bootstrap.go
 	go-bindata -pkg bootstrap -o bootstrap/bootstrap.go bootstrap/
-
 
 build: authors clean build-linux-amd64 build-darwin-amd64 build-freebsd-amd64 build-windows-amd64
 
@@ -19,6 +26,7 @@ authors:
 
 clean:
 	rm -rf bin/*
+	rm -rf bootstrap/bootstrap.go
 
 gofmt:
 	gofmt -w ./apis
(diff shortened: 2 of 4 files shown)


paulczar:
probably need to break out install into different doc, but I have to run and want to at least push up what I have for comment.

Makefile
```diff
@@ -2,20 +2,23 @@ PKGS=$(shell go list ./... | grep -v /vendor)
 SHELL_IMAGE=golang:1.8.3
 GIT_SHA=$(shell git rev-parse --verify HEAD)
 VERSION=$(shell cat VERSION)
+PWD=$(shell pwd)
 
 default: bindata compile
+
 compile:
-	@go build -o ${GOPATH}/bin/kubicorn -ldflags "-X github.com/kris-nova/kubicorn/cmd.GitSha=${GIT_SHA} -X github.com/kris-nova/kubicorn/cmd.Version=${VERSION}" main.go
+	go build -o bin/kubicorn -ldflags "-X github.com/kris-nova/kubicorn/cmd.GitSha=${GIT_SHA} -X github.com/kris-nova/kubicorn/cmd.Version=${VERSION}" main.go
```
kris-nova: The reason I wanted it to go to `$GOPATH/bin` was so that it would land in my `$PATH`. 

(I have `PATH=$PATH:$GOPATH/bin`)

I am fine with this thought, just giving context

.gitignore
```diff
@@ -1,4 +1,5 @@
 _state*
 .idea*
 .DS_Store
-bootstrap/bootstrap.go
\ No newline at end of file
+bin/*
+bootstrap/bootstrap.go
```
kris-nova: Can we remove bootstrap.go please?

See #51 for context but TLDR:

We need bootstrap.go in so that `go get` works (makes sense). We need to come up with another approach because I really hate the idea of having a generated file in the repo. But for now please remove? Thoughts?

README.md
```diff
@@ -14,15 +14,35 @@ With `kubicorn` a user can create new clusters, modify and scale them, and take
...
+
+### Quickstart
+
+Assuming you have a local Go environment The Simplest method to install is to just run:
+
 ```
 $ go get github.com/kris-nova/kubicorn
 ```
 
+### Not so Quickstart
```
kris-nova: Can we *only* have the quickstart on the main readme, and then add a link to an in depth building.md in `/doc`?

I want the main readme to be short/sweet and I have a feeling we are going to have to start documenting the Makefile anyway down the road. What are your thoughts here? I am kind of on the fence.

kris-nova: Just saw your comment about pushing this to another doc 😆 

I vote doing that in another PR and letting this one in.

paulczar: Yeah agreed!  I'll work up a separate install doc tomorrow.

Makefile
paulczar: yeah, I'm just not a huge fan of a make putting files outside of the source directory outside of a `make install`.  but I'm sure `$GOPATH/bin` is fine, I think I just caught this with the other dirs I was changing.

.gitignore
paulczar: it's already back in the `master` branch right?  this will just make sure people don't add changes to it to their commits if that do a `git add .` rather than remove it again  ...  unless we want that behavior?

README.md
kris-nova: Sounds good thank you!!!

.gitignore
kris-nova: Yep good call, let's leave this as is

Makefile
kris-nova: sounds good

k4k: @kris-nova Would it be acceptable here to add a `make install` step which places the built files in `${GOPATH}/bin/kubicorn`? I agree that `make` should be equivalent to a `make build`. A `make all` step could also be added to replace the current behavior of plain `make` if you think `make build install` would be annoying.

kris-nova: Not annoying - I like this plan.

1. `make` compile only into `./bin`
2. `make install` move compiled files from `./bin` -> `$GOPATH/bin` (we should probably use `go install here`)
3. `make all` wraps the above two targets

paulczar:
updated the PR with feedback and rebased to remove some merge conflicts. :crossed_fingers:

PR #55, opened 2017-07-16
Support MFA and roles when doing auth

jbeda:
We set up our AWS so that the default account has little/zero permissions and we do everything through roles across accounts.  We also require MFA in a lot of places.  This magic incantation when creating the session makes the SDK act more like the aws command line tools.  

You can do something like `export AWS_DEFAULT_PROFILE=dev-joe` to set the profile to the one that you want to use and now kubicorn will pick it up.


kris-nova:
LGTM thanks @jbeda

PR #81, opened 2017-07-18
Fix #76 -- Prompt for ssh key passphrase

### LINKED ISSUE #76: ssh: cannot decode encrypted private keys
```
kubicorn getconfig -n attractive-badger -v4
2017-07-17T11:44:46-04:00 [✔]  Selected [fs] state store
2017-07-17T11:44:46-04:00 [✔]  Loaded cluster: attractive-badger
2017-07-17T11:44:46-04:00 [✔]  Init Cluster
2017-07-17T11:44:46-04:00 [▶]  Running preprocessors
2017-07-17T11:44:46-04:00 [✖]  ssh: cannot decode encrypted private keys
```

I believe this is related to: https://github.com/kris-nova/klone/issues/13

k4k:
This should be refactored to be better but works around the issue in #76 for now

What I would ultimately like to do is inspect the key to determine if it's password protected, attempt to read from ssh-agent and then fall back to ask for password. I need to read more `crypto/ssh` stuff first.

### DIFF
--- a/cutil/kubeconfig/kubeconfig.go
+++ b/cutil/kubeconfig/kubeconfig.go
@@ -7,6 +7,7 @@ import (
 	"github.com/kris-nova/kubicorn/logger"
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/terminal"
 	"io/ioutil"
 	"os"
 	"strings"
@@ -32,7 +33,8 @@ func GetConfig(existing *cluster.Cluster) error {
 	if err != nil {
 		return err
 	}
-	signer, err := ssh.ParsePrivateKey(pemBytes)
+
+	signer, err := GetSigner(pemBytes)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cutil/kubeconfig/kubeconfig.go
```diff
@@ -32,7 +33,14 @@ func GetConfig(existing *cluster.Cluster) error {
 	if err != nil {
 		return err
 	}
-	signer, err := ssh.ParsePrivateKey(pemBytes)
+
```
kris-nova: Can we have the code try it without a passphrase first? Then if we get a passphrase err prompt for one? 

Also we will need a way to disable all together via an environmental variable. User's will be wrapping the apply in automation scripts, and we can't leave them hanging (pun intended)

k4k:
I've refactored to assume no passphrase first and only prompt on `err`. Still no bypass method for library use. I would like to add support for `[]ssh.AuthMethod {SSHAgent()}` to help with library use cases instead of (or in addition to?) using an environment variable but I still have some figuring out to do there. I'll need to more intelligently structure how I'm checking the key for that to work.

@kris-nova If you'd rather just get this merged I add support for a "NOPASSWD" environment variable which could be used to short circuit the attempt to use `ssh.ParsePrivateKeyWithPassphrase()`

kris-nova:
closed #76

kris-nova:
Opened up #91 for the next step

Issue #68, opened 2017-07-17
Implement CI/CD Testing


kris-nova:
I have a credit card, if you have skills and opinions!

### Goal

run `make test` for every PR and only allow merging on exit `0`

jmcarp:
Do you have a preferred CI service?

kris-nova:
@jmcarp just now seeing this - looks like @paulczar was able to add travis support for us. I have used it in the past and want to start there.

Do you have experience with any other CI/CD tools that you think would be helpful? I am asking because given the events of the bug in kubeadm today, I am realizing we REALLY need a rock solid build

PR #85, opened 2017-07-19
add support for travis.ci testing

paulczar:
This should resolve issue #68 by adding travis.ci support.  It's free for opensource
so should be easily enabled by connecting travis.ci with the repo by somebody who has access ( @kris-nova ) to connect via oauth and authorize travis to the repo.

I figured it would be worth doing both `make test` and `make compile` to ensure both steps work in a fairly blank go environment on PRs/Merges.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: go
+
+go:
+  - 1.8.3
+
+script:
+  - make test
+  - make compile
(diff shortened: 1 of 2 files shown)


kris-nova:
@paulczar this is fabulous! I might need some help setting it up, but will give it a go! I will let you know if I get snagged.

Cheers

Issue #91, opened 2017-07-20
Bypass password auth for SSH keys by global env var


kris-nova:
We need to have an optional environmental variable to disable ALL password interactive auth (for our friends in CI/CD land)

### Success Criteria

1. The environmental variable is documented, and follows the standard `KUBICORN_` prefix. (You can see examples in any of the cobra commands)
2. The environmental variable is *always* called using `os.Getenv` and never cached
3. Please define the new variable here https://github.com/kris-nova/kubicorn/blob/master/docs/envar.md

kris-nova:
@k4k FYI

kris-nova:
closed with #299

PR #102, opened 2017-07-22
Use SSH Agent if present

paulczar:
If ssh agent is running and SSH_AUTH_SOCK is set we
should attempt to use it rather than try to load the keys
this way we can bypass need to ask for passphrase.

This might address #91


k4k:
This is great! Definitely something we wanted to support. I'm not sure that Travis will have SSH agent running so I'm not sure if it solves #.91 but someone with more Travis experience may know better.

k4k:
I'm having trouble making this work. I have my ssh keys added to ssh-agent but I'm still being prompted for a password. @paulczar what platforms did you test this on? I'm wondering if there's something strange with my environment causing `sshAgent()` to fail.

kris-nova:
Out of town at the moment - will check out ASAP thank you so much for the PR

kris-nova:
The code looks sound, merging to master and we can iterate if needed. @k4k can we open an issue to track your agent not working and track there?

kris-nova:
See issue #104 regarding standardizing an scp package

PR #101, opened 2017-07-22
Rebase to change incorrect author info (take 2)

k4k:
This fixes my email address across the repo history. Upon inspection of the commit history with `git log --author="ted@xy0.org" appears correct. Please double check that I did not clobber any other commits.

For reference by future users, here was the command that saved the day:

```
git filter-branch --env-filter 'if [ "$GIT_AUTHOR_EMAIL" = "ted@mailchimp.com" ]; then \
GIT_AUTHOR_EMAIL="ted@xy0.org"; \
GIT_AUTHOR_NAME="Ted Wood"; \
GIT_COMMITTER_EMAIL="$GIT_AUTHOR_EMAIL"; \
GIT_COMMITTER_NAME="$GIT_AUTHOR_NAME"; fi' -- --all
```

PR #114, opened 2017-08-01
Updating Authors file for #kubicorn

kris-nova:


AUTHORS
```diff
@@ -1,8 +1,12 @@
...
+Igor Vuk <parcijala@gmail.com>
 Joe Beda <joe.github@bedafamily.com>
 Joshua Carp <jm.carp@gmail.com>
+@Kris__Nova <kris@nivenly.com>
 Kris Nova <kris@nivenly.com>
+Marko Mudrinić <mudrinic.mare@gmail.com>
 Michael Hausenblas <michael.hausenblas@gmail.com>
 Paul Czarkowski <username.taken@gmail.com>
 Ted Wood <ted@mailchimp.com>
+Ted Wood <ted@xy0.org>
```
k4k: ...I think I see what happened: https://stackoverflow.com/a/40552998/6177740

I'm working on cleaning this up and will submit a PR when I've done it. Sorry for the mess.

k4k: OK... nevermind, sorry to bug you on this. The only thing I can do here is revert #101 which is pointless because as soon as I make another commit I'll just wind up back in here with two email addresses again 😢  Since it looks like you're in there twice anyway, I'll look to find alternative ways of generating the authors file and PR it if I find something.

PR #78, opened 2017-07-17
Build kubicorn from source (issue #77)

arashkaffamanesh:
Please add additional instruction to build kubicorn from source #77

PR #116, opened 2017-08-01
Docker build script

MyDigitalLife:
I created a script that performs a make with docker. My work machine is a Windows 10 machine most of the time so make files are not supported by default. That's why I created a small docker image and a .sh script to build and run a docker image that does the build process for me.

Simple and easy. Just run build.sh in the docker folder if you have docker installed. I thought it would be nice to include this in the project. I tested it on windows 10 and on Ubuntu and both work like a charm.


k4k:
@kris-nova Would you be opposed to standardizing the build on docker? It would provide some added benefits around build environment sanitization and standards.

docker/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+echo Building container
+docker build -t gobuilder "$(pwd)"
+echo Running make script
+docker run --rm -v "/$(pwd)/.."://usr/src/myapp -w //usr/src/myapp gobuilder make
```
kris-nova: Can we please mount to `/go/src/github.com/kris-nova/kubicorn` and have `$GOPATH=/go`?

kris-nova:
@k4k I am all for having a docker build,

although I wouldn't want it to replace local build logic, but rather live next to it. Does that make sense at all? If you think it's a good idea please PR it and/or open an issue and we can go from there!

MyDigitalLife:
I agree with @kris-nova that the docker build should not replace the existing process. Not everybody want to install docker or know what it is/does. 

I've added a remove image option and a verbose option. Its now almost completely silent by default and outputs everything when adding -v or --verbose. I can also add the documentation from #78 and add some extra text about the docker build.

docker/build.sh
MyDigitalLife: This has been updated. And the GOPATH variable is /go by default (https://github.com/docker-library/golang/blob/2a15dfff04accfd31c2a45b3bb7423aa86aa2d60/1.8/jessie/Dockerfile line 45).

docker/build.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+VERBOSE=FALSE
+VERBOSE_DOCKER_RUN=""
+VERBOSE_DOCKER_BUILD="-q"
+REMOVE_IMAGE=FALSE
+
+docker >/dev/null 2>&1 || { echo >&2 "Docker is required but it's not installed. Aborting."; exit 1; }
```
k4k: Minor nit-pick but I don't like that we're blindly running whatever `docker` is in $PATH here just to check that it's available. Can we just `which docker > /dev/null || {...}` here instead?

kris-nova:
LGTM - thanks everyone for your help

Issue #88, opened 2017-07-19
Apache 2 License Headers


kris-nova:
In order for us to be truly Apache 2 compliant we need headers on top of **all** source code files

So can we please build out support for the following:

`make headers` that will run an idempotent "fix" for all files in the repository (sans vendor).
`make check-headers` that will run a check to validate all files have correct license headers. We will eventually want this to be called from `make test` and fail if a license has the wrong or missing header.

Every header should reference `The Kubicorn Authors` which is the `/AUTHORS` file that is automatically generated every release

PR #117, opened 2017-08-01
makefile: check-header implementation

xmudrii:
Partially addresses #88 

This implements `check-header` command for Makefile. Running `make check-header` returns the following:
```
incorrect license header: ./bootstrap/bootstrap.go
incorrect license header: ./bootstrap/inject.go
incorrect license header: ./cutil/hang/ratio.go
incorrect license header: ./cutil/kubeadm/token.go
incorrect license header: ./cutil/initapi/init.go
incorrect license header: ./cutil/initapi/ssh.go
incorrect license header: ./cutil/initapi/validate.go
...
```

It uses external script, located at the `./scripts` directory, but I think it's easiest and best? way to do this.
If you have anything against this way, I'll revisit this PR :+1:

### DIFF
--- a/scripts/check-header.sh
+++ b/scripts/check-header.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+read -r -d '' EXPECTED <<EOF
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


scripts/check-header.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
```
kris-nova: This is going to sound silly, but we need a header in this file.. 😉

PR #118, opened 2017-08-01
Tweak make authorsfile script and remove duplicate authors

alexlehner:
👋  I'm new here! I saw a tiny thing I may be able to help with. Feel free to ignore :) 

When generating the AUTHORS file the Makefile relies on a git log that may or may not be representative of the valuable data. It's very easy to omit configuration of the git config which easily makes for duplicate authors.

This change tweaks the output of the sorted git log, pipes the values into egrep that relies on an inverse filter that's easily extended. This change should not be seen as a long-term solution as the regex could quickly grow out of hand.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -3,10 +3,8 @@ Eric Hole <eric.hole@gmail.com>
 Igor Vuk <parcijala@gmail.com>
 Joe Beda <joe.github@bedafamily.com>
 Joshua Carp <jm.carp@gmail.com>
-@Kris__Nova <kris@nivenly.com>
 Kris Nova <kris@nivenly.com>
 Marko Mudrinić <mudrinic.mare@gmail.com>
 Michael Hausenblas <michael.hausenblas@gmail.com>
 Paul Czarkowski <username.taken@gmail.com>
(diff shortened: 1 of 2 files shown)


kris-nova:
Hi @alexlehner! Welcome to the project! Please feel free to contribute any way you think you can help, and feel free to add issues to the back log or take a crack at coding them.

kris-nova:
LGTM

We might want to come up with a way of doing the dupe check on the name string, and not hard coding things to ignore - but this looks good for now! Thanks!

alexlehner:
@kris-nova Totally agree that dupe checking by name sounds mucho better. Where I ran into trouble was figuring out programmatically which one was _correct_. For instance @k4k has two valid email addresses, I didn't know of way to do that without some kind of white/black list. We could go with the first (or last) but that's not very transparent to folks casually viewing the AUTHORS file. 

Plus, we'd need to check dupes on both emails and first names since the repo has a git history of `@Kris__Nova` and `Kris Nova`. 

Suddenly a manually managed AUTHORS file doesn't sound so bad. 😆 

Thank you again for the welcome and super fast merge!

PR #120, opened 2017-08-02
makefile: add headers command

### LINKED ISSUE #88: Apache 2 License Headers

xmudrii:
Fixes #88 

This is my try on `headers` command. It adds headers to all Go files.
Disclaimer: I'm big sed noob, so forgive me if it's awful solution. If it's that bad, I can try digging better one.


kris-nova:
Can we please have travis check the headers using the make target you introduced in an earlier PR? I would like to see this PR pass that check if at all possible.

xmudrii:
@kris-nova done, let's see will travis :green_apple:

kris-nova:
LGTM thank you!

PR #123, opened 2017-08-02
scripts: make check-header return exit code

xmudrii:
We've added `make check-header` to CI so license headers are tested on every PR/commit.

The problem is that in current implementation it doesn't return error code so it will pass every time.
I've changed this to return code 1 if incorrect header is found. In that case CI will fail.


kris-nova:
Perfect - thank you

PR #125, opened 2017-08-03
Added help command to script to document flags

MyDigitalLife:
A small addition to the Docker build script so all available flags are documented an can be called using a -h or --help flag.


kris-nova:
LGTM

Issue #146, opened 2017-08-04
Add example script for digital ocean


kris-nova:
We offer example scripts for users who want to use kubicorn as a library. We have one for AWS here
https://github.com/kris-nova/kubicorn/blob/master/examples/aws.go

Can we please make one for digital ocean?

MyDigitalLife:
I've added a digital ocean example and created a pull request #148. I have however not been able to test this because i could not find any documentation no how to use this. If you could point me in the right direction i would test it fully.

kris-nova:
You should just be able to `go run` the file

MyDigitalLife:
Looks like there is still a problem here. When running `go run basiccluster.go` i get the following output:

````
2017-08-05T22:48:15+02:00 [✔]  Init Cluster
2017-08-05T22:48:15+02:00 [▶]  Running preprocessors
2017-08-05T22:48:15+02:00 [▶]  ssh.Expected
2017-08-05T22:48:15+02:00 [▶]  ssh.Render
2017-08-05T22:48:15+02:00 [▶]  droplet.Expected
2017-08-05T22:48:15+02:00 [▶]  droplet.Render
2017-08-05T22:48:15+02:00 [▶]  droplet.Expected
2017-08-05T22:48:15+02:00 [▶]  droplet.Render
2017-08-05T22:48:15+02:00 [▶]  ssh.Actual
2017-08-05T22:48:15+02:00 [▶]  ssh.Render
2017-08-05T22:48:15+02:00 [▶]  droplet.Actual
2017-08-05T22:48:15+02:00 [▶]  droplet.Render
2017-08-05T22:48:15+02:00 [▶]  droplet.Actual
2017-08-05T22:48:16+02:00 [▶]  droplet.Render
2017-08-05T22:48:16+02:00 [▶]  ssh.Expected
2017-08-05T22:48:16+02:00 [▶]  Using cached ssh [expected]
2017-08-05T22:48:16+02:00 [▶]  ssh.Actual
2017-08-05T22:48:16+02:00 [▶]  Using cached ssh [actual]
2017-08-05T22:48:16+02:00 [▶]  droplet.Apply
2017-08-05T22:48:16+02:00 [▶]  Actual   : 3aa0d9d5c8e4d36d844617b475d278fc
2017-08-05T22:48:16+02:00 [▶]  Expected : 76476dc182047a11b96afb03f50093aa
2017-08-05T22:48:16+02:00 [✔]  Created SSH Key [11664620]
[...]

PR #148, opened 2017-08-04
Added digital ocean example 

MyDigitalLife:
As asked in #146 i created a example for digital ocean. I have not been able to test this because there isn't really any documentation on how to use this.


looztra:
the digitalocean profile and code is already available, see https://www.nivenly.com/kubernetes-on-digital-ocean-with-encrypted-vpn-service-mesh/

MyDigitalLife:
Hmm the Travis build is failing because there are now 2 files with the main function in the same directory.  I can edit the Makefile to exclude the examples but that would mean that the files are never tested so it might not be the best solution. I think it might be a better idea to sub folder the examples (if that fixes it). So it would be something like this:
```
examples/digitalocean/example1.go
examples/aws/example2.go
```

What do you think @kris-nova

kris-nova:
I vote nested folders `examples/$cloud/$example.go`

examples/digitalocean/basiccluster.go
```diff
@@ -0,0 +1,150 @@
...
+		panic(err.Error())
+	}
+	err = reconciler.Destroy()
+	if err != nil {
+		panic(err.Error())
+	}
+}
+
+func getCluster(name string) *cluster.Cluster {
+	return &cluster.Cluster{
```
kris-nova: Can we make the profile match this one please? https://github.com/kris-nova/kubicorn/blob/master/profiles/simpledigitalocean.go#L23

PR #163, opened 2017-08-05
kubeconfig: create .kube dir if it doesn't exist

xmudrii:
Fixes issue that if you don't have `~/.kube` dir on local machine it will fail copying kubeconfig from the remote.


kris-nova:
Can we please add a test that checks this both happy and sad?

Issue #161, opened 2017-08-05
The readme is a table - please help


kris-nova:
Right now our readme is a table, and I feel like that is probably the most unattractive thing ever.

Please PR something that makes the README.md feel warm, inviting, and smart.

PR #164, opened 2017-08-05
docs: improve readme

xmudrii:
Proposal to use center logo rather than using table. I don't like that table that much.

Suggestions appreciated


kris-nova:
Closes #161

kris-nova:
LGTM

Issue #142, opened 2017-08-04
Linting


kris-nova:
Despite my desire to keep our pull request experience light and easy, I think having some basic linting would be helpful.

Can somebody set up a lint check in our `Makefile` and have travis run it?

k4k:
In looking around at options for this, it seems linting a Makefile is more difficult than it may appear on the surface. Can we try to identify some specific goals for the linter? From looking at the make file now it looks like this would be a good start:

* Whitespaces
* Unused variables
* `make all --dry-run`

xmudrii:
‎I believe @kris-nova is talking about linting Go code rather than linting Makefile.

It could be done same as `gofmt`, just to make sure user actually has golint installed (do `go get`)

k4k:
@xmudrii Go lint is already being run by virtue of runnin `make compile` in the existing tests. Do we want to split these lint checks out separately from the build checks?

xmudrii:
‎That's not lint, take a look at this https://github.com/golang/lint

MyDigitalLife:
Fixed with #170. We got about 240 warnings about name formatting, adding comments and not using under scores in names. I can easily fix the name formatting but the comments is probably something @kris-nova should fix seeing she wrote most of the existing code and knows best what this are and do.

xmudrii:
Partially fixed with #187

kris-nova:
Update? Are we linting yet? I haven't gone through and written a ton of code docs so I assume we still have at least 1 more step to go right?

MyDigitalLife:
The docs is definitely still a open issue and with the code of the last few days there are probably a few more now. I would love if we can handle this ASAP and then add it to Travis. This way it will be mandatory for the project to have no linting errors.

paulczar:
Here's a PR to enable bash script linting: https://github.com/kris-nova/kubicorn/pull/340

PR #170, opened 2017-08-05
Added golint to makefile

MyDigitalLife:
Added the linting functionality as requested in #142.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,6 +76,10 @@ shell:
 	-v ${PWD}:/go/src/github.com/kris-nova/kubicorn \
 	--rm ${SHELL_IMAGE} /bin/bash
 
+lint:
+    which golint > /dev/null || go get -u github.com/golang/lint/golint...
+    golint $(PKGS)
+
 .PHONY: test
 test:
 	go test -timeout 20m -v $(PKGS)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -76,6 +76,11 @@ shell:
 	-v ${PWD}:/go/src/github.com/kris-nova/kubicorn \
 	--rm ${SHELL_IMAGE} /bin/bash
 
+lint:
+	which golint > /dev/null || go get -u github.com/golang/lint/golint...
```
xmudrii: There are three dots on the end which shouldn't be there. You can just use `go get -u github.com/golang/lint/golint`

MyDigitalLife: I took this from the example on line 23, it also has the three dots. Can remove them because it looks like it doesn't really affect the functionality.

xmudrii: It's not the same situation. If you take a closer look, you'll see that you have something such as `go-bindata/...`, which expands and installs all the packages under that directory.
That's not the case with `golint`. Inside `github.com/golang/lint/golint` you don't have any other package beside `golint`, so you just need to install that one. 
I hope you understand better how it works. :)

MyDigitalLife: Ah oke, got it!

xmudrii:
Thank you very much @MyDigitalLife :tada:

PR #172, opened 2017-08-06
Integrate visual indicator that the cluster destroy action is in progress.

metmajer:
Resolves https://github.com/kris-nova/kubicorn/issues/27.

Above issue seemed simple enough for my first contribution in Golang. @kris-nova: please provide feedback on code quality. Would you also suggest to factorize the anonymous function that takes care of printing the visual cue "." into a package? If so, where would you see good fit?

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -101,8 +101,35 @@ func RunDelete(options *DeleteOptions) error {
 		return fmt.Errorf("Unable to init reconciler: %v", err)
 	}
 
-	if err := reconciler.Destroy(); err != nil {
-		return fmt.Errorf("Unable to destroy resources for cluster [%s]: %v", options.Name, err)
+	done := make(chan bool)
+	message := make(chan error)
+
+	go func() {
+		message <- reconciler.Destroy()
+	}()
+
+	go func(description string, symbol string, c chan bool) {
+		if description != "" {
+			fmt.Print(description)
+		}
+
+		for {
+			select {
+			case quit := <-c:
+				if quit {
+					return
+				}
+			default:
+				time.Sleep(200 * time.Millisecond)
+				fmt.Print(symbol)
+			}
+		}
+	}(fmt.Sprintf("Destroying resources for cluster [%s]:\n", options.Name), ".", done)
+
+	err := <-message;
+	done <- true
+	if err != nil {
+		return errors.Errorf("Unable to destroy resources for cluster [%s]: %v", options.Name, err)
 	}
 
 	if options.Purge {
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"os"
+	"time"
 )
 
 // deleteCmd represents the delete command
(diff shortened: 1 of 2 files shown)


cmd/delete.go
```diff
@@ -101,8 +102,35 @@ func RunDelete(options *DeleteOptions) error {
 		return fmt.Errorf("Unable to init reconciler: %v", err)
 	}
 
-	if err := reconciler.Destroy(); err != nil {
-		return fmt.Errorf("Unable to destroy resources for cluster [%s]: %v", options.Name, err)
+	done := make(chan bool)
+	message := make(chan error)
```
kris-nova: Can we stick with the convention of calling errors `err` please?

cmd/delete.go
```diff
@@ -101,8 +102,35 @@ func RunDelete(options *DeleteOptions) error {
...
+
+	go func() {
+		message <- reconciler.Destroy()
+	}()
+
+	go func(description string, symbol string, c chan bool) {
+		if description != "" {
+			fmt.Print(description)
```
kris-nova: Can we use `logger.Info` instead of `fmt.Print`? It's soops important all of our output goes through the same place so we can manage things globally.

cmd/delete.go
```diff
@@ -101,8 +102,35 @@ func RunDelete(options *DeleteOptions) error {
...
+
+		for {
+			select {
+			case quit := <-c:
+				if quit {
+					return
+				}
+			default:
+				time.Sleep(200 * time.Millisecond)
+				fmt.Print(symbol)
```
kris-nova: Please funnel this through the `logger` package as described here https://github.com/kris-nova/kubicorn/pull/172/files#r131545566

kris-nova:
While I think pulling this into a package so we can have a fun wait system for other parts of the program is necessary, I think coding it into one place (like this PR) is a good and meaningful starting point to see how it goes moving forward.

My vote:

1. Check out my comments on the PR and merge this in
2. Open up an issue to create a `sleeper` package or something similar that will perform this same action, but with arbitrary input.

Thoughts?

cmd/delete.go
metmajer: Sure! Should we follow the same pattern with `done`, i.e. `donechan`?

cmd/delete.go
metmajer: Thanks for your explanations and guidance, Kris! Changing this to a `logger.Info()` seems sensible.

metmajer:
Makes sense! Will create the issue once we've got this merged.

cmd/delete.go
kris-nova: Yes please :smile:

cmd/delete.go
kris-nova: Okay great. Will `logger.Info` work?

metmajer: I've just seen that `logger.Level` is initialized to 2, so users wouldn't see the process activity indicator unless they move to at least `warning` level. Therefore, `logger.Always` seems to be more applicable.

metmajer: I've just seen that `logger.Level` is initialized to 2, so users wouldn't see the process activity indicator unless they move to at least warning level. `logger.Always` isn't the best option either because of the hard-wired ✿ label. What are your thoughts on introducing a function `logger.Log` that ignores `logger.Level` and that does not apply any label?

metmajer: @kris-nova, I must have somehow missed your edits in your upper message. Seems like we're on the same page now 👍

kris-nova:
Gorgeous - LGTM

kris-nova:
And congrats on your first Go PR!! :smile:

PR #174, opened 2017-08-06
Fixed problem with asking for password on windows.

MyDigitalLife:
When asking for the password it was giving errors on windows without asking for a password. Fixed thanks to @xmudrii.


kris-nova:
Samesies

PR #186, opened 2017-08-07
Bash and Zsh completion for flags

### LINKED ISSUE #26: Autocomplete state names
Should be able to tab complete the state names from the list of states in _state using bash/zsh completion

### LINKED ISSUE #128: Bash and Zsh completion for commands
I would like to see bash and zsh completions for commands. With cobra, it's easy to get started and I should have PR with basic completions quite soon.

Lately, we need to add completions for resources. There we need to write a few functions to get stuff from `_state` and relevant locations.

xmudrii:
Fixes #128 
Fixes #26 

Downsides:
* Current version supports only `./_state` or using environmental variables. This should not be a big thing, even if it stays.
* Profiles are typed in as constant. We will fix this once we better know how we will manage profiles. We will probably need an command to list all profiles.

Available for `--name` and `--profile` flags.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -52,11 +52,22 @@ type ApplyOptions struct {
 
 var ao = &ApplyOptions{}
 
-func init() {
-	RootCmd.AddCommand(applyCmd)
+func init() {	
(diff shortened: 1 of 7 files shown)


kris-nova:
LGTM thanks

PR #187, opened 2017-08-07
Renamed variable names to fix lint warnings

MyDigitalLife:
There are still 196 warnings after this. But these are all just missing comments.


xmudrii:
Thank you very much @MyDigitalLife

Issue #178, opened 2017-08-06
Add OpenVPN bootstrap script for DigitalOcean


xmudrii:
Due to meshbird being not so good VPN solution (see issue #171), we need to offer multiple solutions.
OpenVPN is popular solution but we need to make it work in mesh mode so we actually get private networking. This [DO tutorial](https://www.digitalocean.com/community/tutorials/how-to-secure-traffic-between-vps-using-openvpn) and [this](https://www.digitalocean.com/community/tutorials/how-to-set-up-an-openvpn-server-on-ubuntu-16-04) could be a good starting point.

We need to see will it be okay for our need, create bootstrap script and test it.

Also see #177 for WeaveMesh.

kris-nova:
This is prime bash hacking territory if we have any bash hackers among us!

prateekgogia:
Hi, I can take a stab at it. Does it really have to be in bash ? I have some ideas to run openVPN in Docker containers with host networking and deploy using Kubernetes itself. Don't have the complete details yet, but if you think it will be something useful I can try to implement this.

xmudrii:
Personally, I think we should stay with pure bash if that's possible. I believe that it could be easier for users to later change it per their needed with bash rather than with Docker. Also, we shouldn't mix implementations (e.g. one docker, one bash, another one something other).

xmudrii:
I'm starting with this one. Nobody else started in meanwhile?

Issue #177, opened 2017-08-06
Add WeaveMesh VPN bootstrap script for DigitalOcean


xmudrii:
Due to `meshbird` being not so good VPN solution (see issue #171), we need to offer multiple solutions.
WeaveMesh seems like a good solution - https://github.com/weaveworks/mesh

We need to see will it be okay for our need, create bootstrap script and test it.

Also see #178 for OpenVPN issue.

kris-nova:
Any research here from anyone? Will this solve our problem?

dpraburaj:
My understanding of k8s' networking(and networking in general) is very rudimentary but wonder if enabling WeaveNet's encryption(https://www.weave.works/docs/net/latest/concepts/encryption/) would make a VPN mesh like Meshbird unnecessary. 

As far as I understand, the control plane communication happens securely in kubeadm clusters by default and enabling weave net encryption(if one's using weave for overlay) would secure the data
data plane communication.

xmudrii:
Duplicate of #415 

@dpraburaj I'll close this issue, as we have new one for this, #415
Basically, the biggest reason why we have VPN for DO is because Private Networking is not really "private".
Every Droplet from that Datacenter (even Droplets we don't own) can send/access our Droplets. So we wanted to increase security and we implemented VPNs.
However, with #407, we increased security by implementing Cloud Firewalls, so this is no a problem anymore.

More details is also available in issue #415

fntlnz:
Another valid alternative could be wireguard: https://www.wireguard.com/

xmudrii:
@fntlnz It looks good, however it looks like WeaveNet is better optimized for running on Kubernetes stack. I'll look around, this could be useful as well -- it seems like it's very easy to set up and configure it. Thanks a lot for the recommendation.

PR #192, opened 2017-08-07
Decouple meshbird

MyDigitalLife:
This is a first step for #177 and #178. With this pull request the VPN is setup with a different script that can be replaced. Next step will be to write the bash scripts and add WeaveMesh and OpenVPN.


apis/cluster/serverpool.go
```diff
@@ -35,6 +35,7 @@ type ServerPool struct {
 	Image             string
 	Size              string
 	BootstrapScript   string
+	VPNScript         string
```
kris-nova: So this PR is *fantastic* but I want to make sure we treat our API very delicately. Even though this very well might get refactored completely moving forward, we still should consider our changes carefully.

I love the idea of having the VPN logic pulled out into different scripts, and then appending them into the `customdata` in the droplet resource (what you have done here) but I am wondering if having a new member called `VPNScript` is the right approach.

Would it be better if we changed `bootstrapScript` -> `bootstrapScripts` and made it of type `[]string` instead?

We could then guarantee they are ordered correctly (using a proper `for` loop and index) when we append them.

This would be an easy change, and would allow us to add many scripts at runtime, instead of building in new members for every new type of script. Thoughts?

If we changed the API we would have to refactored the AWS implementation as well. It shouldn't be hard, and at this point I am *okay* with breaking the API because we aren't advertising this to be used yet.

MyDigitalLife: I really like this idea! Its a lot more logical and makes it way easier when we add all kinds of other script. Less code, more functionality.

apis/cluster/serverpool.go
```diff
@@ -34,7 +36,20 @@ type ServerPool struct {
 	Name              string
 	Image             string
 	Size              string
-	BootstrapScript   string
+	BootstrapScripts  []string
 	Subnets           []*Subnet
 	Firewalls         []*Firewall
 }
+
+func BuildServerPoolScript(serverPool *ServerPool) ([]byte, error) {
```
kris-nova: Can we decouple this function from the API? Instead of passing in a serverPool can we just pass in a slice of strings?

That way we just have a generic append function?

kris-nova:
One small nit pick then we can merge, thanks for helping!

kris-nova:
LGTM

PR #193, opened 2017-08-07
godo: parse ssh errors

xmudrii:
This should fix the error where even in cases such as corrupted SSH key, user would get key not existing instead of appropriate error.


kris-nova:
LGTM

PR #194, opened 2017-08-07
Migrate progress indicator code into package 'task'.

metmajer:
Resolves https://github.com/kris-nova/kubicorn/issues/190.

This pull request adds a package `cutil/task` with convenience methods to annotate long running tasks with an arbitrary description and a periodic sequence of symbols.

Looking forward to your feedback, @kris-nova and @xmudrii 👍


xmudrii:
Hi @metmajer thank you very much for this.

To me it look great, but I'll leave it on @kris-nova for detailed review as she's much better friend with channels than me. :)

kris-nova:
Can we get happy/sad tests here please @metmajer? I think @xmudrii can explain

xmudrii:
‎I'm on mobile and already in the bed - but I should be able to explain in short. 

We want to make two tests: Happy test, which will ensure that something works as we want.
Also we want Sad test, which ensures that action incorrect request is handled as it should be.

For the example, you can take a look at this https://github.com/kris-nova/kubicorn/blob/master/cloud/digitalocean/godoSdk/sdk_test.go
In case we provided DO token and got Godo Client it happy test will pass.
Sad tests verify that if we didn't provide token, function will return an error. In case it indeed returns error, Sad test will pass.

I hope you will understand. If you need any help, I'll try to help as much as I can.

metmajer:
Thanks @xmudrii, I am good with the information I have. With how code is written in the `logger` package at the moment, the best choice I have to unit test the `task` package is to assert the expected strings to be written to `stdout`. However, this is not a desirable strategy.

A better one would be to adapt our `logger` implementations to support a `type io.Writer`. The argument would be an `os.Stdout` for production code and a `bytes.Buffer` for test code. This way, we could more easily test our `logger` implementations and whatever depends on them.

Design-wise, I'd suggest to expect a `type io.Writer` as the last argument to the variadic `a ...interface{}` parameter. I have verified that [github.com/fatih/color](https://github.com/fatih/color) supports passing of `io.Writer` (and we're already using this concept for Fabulous).

@kris-nova and @xmudrii: What do you think? I'd take care of implementation and tests if you consider this as a worthy improvement.

kris-nova:
@metmajer this seems brilliant, make the `writer` changes to logger 👍 

I just want to make sure we are testing our task package for timeouts both happy and sad. Feel free to code in whatever improvements into `logger` you think would help. Your design seems sane. 😄

metmajer:
@kris-nova: sure, I typically follow a test-driven approach, so having tests present makes most sense. As I've gotten more comfortable with Golang over the past days, I'll make sure to add tests to all future contributions.

I'll keep this PR open for a while until I have adapted the `logger` package, which I want to handle in a separate issue first.

kris-nova:
@metmajer you are quite brilliant, sounds perfect. Let me know how I can help.

metmajer:
This update adds tests for the `task` package. Instead of running a timer in the test (which I didn't consider good design), I've calculated upper and lower bounds of repeated symbols (e.g. '.') and made sure those bounds are met via a regexp construct `\.{lower}{upper}`.

@kris-nova: I've heard your request about about both happy and sad tests, but the only sad paths I could think of were about providing nil values as arguments. Since these tests wouldn't add much value on top of the happy path, I left them out for the moment - but please let me know if you see better sad path candicates.

On a personal note, I am not sure how happy you are with the implementation of `task.RunAnnotated`. It follows the practice we used in `logger.*` where the variadic parameter `a ...interface{}` was used to inject arbitrary types which are then coerced at runtime. This adds some flexibility and cleaner interfaces, but may not be very idomatic to golang being a statically typed language. As an improvement, I removed any uses of the reflection package and turned to simple type assertions instead.

Thoughts?

cutil/task/task.go
```diff
@@ -0,0 +1,75 @@
...
+var (
+	DefaultTicker = time.NewTicker(200 * time.Millisecond)
+)
+
+// annotates a task with a description and a sequence of symbols indicating task activity until it terminates
+func RunAnnotated(task Task, description string, symbol string, options ...interface{}) error {
+	donechan := make(chan bool)
+	errchan := make(chan error)
+
+	logger_ := logger.Log
```
prateekgogia: Please don't use "_" in variable names, Go lint will complain.

cutil/logger/logger.go
```diff
@@ -137,6 +125,20 @@ func Warning(format string, a ...interface{}) {
 	}
 }
 
+func extractLoggerArgs(format string, a ...interface{}) ([]interface{}, io.Writer) {
```
prateekgogia: I was trying to find out if there is any documentation for logging, but couldn't find. Just trying to wrap my head around this, is the expectation is when someone logs using this package, they can pass in an io.writer as the last argument?  Ex -

logger.Log("Created something %s", id, myLogger) 

Looks like we are expecting an io.writer in the end of arguments, which makes it really confusing to use. If we are expecting a writer in each log call, we can accept it as a first argument same as what fmt.Fprintf does in std lib or it can be nil. 
But, I don't thikn passing a writer in each call is a good option. This is something I would pass in for the whole package during init and expect all my logs to go there.

cutil/task/task.go
metmajer: Thanks for the heads up. Reminds me to install `golint`.

cutil/logger/logger.go
metmajer: The idea is to make functions in logger implementations testable by allowing an optional `io.Writer`. Therefore, we don't have to pass it in with every call which keeps the interface simple (if you look at the implementations, you'll see that the `io.Writer` implementation is, in many cases, decided upon by the logger. Examples for this can be found in `cutil/logger/logger_test.go`. This design has been discussed in https://github.com/kris-nova/kubicorn/pull/216.

If you store the `io.Writer` a variable of the `logger` package, you'll have to properly synchronize the access to it, which typically makes the code less simple.

kris-nova: I am going to lean more towards @prateekgogia's suggestion of modeling our functions after `fmt.Fprint()` and explicitly calling out the `io.Writer` as a nilable argument to the function.

Another approach would be to have functions like: `log.Finfo` AND `log.Info` 

However you want to do it is fine with me, but I am worried about introducing a new paradigm that the last argument *should* be a certain type. If it's a type let's call it what it is right?

prateekgogia: Yes, and the reason we should have this right is, this is a logger package and will be used by almost every package and any one writing a new package shouldn't have to go look into the logger package implementation to see, how, stuff works. API should tell him what needs to be passed in to the calls.

kris-nova: Agree

metmajer: I totally get your reasoning and I think that we should do better in the very near future. The sole purpose for supporting an `io.Writer` option in the `a ...interface{}` parameter of each logger function was to enable testing of these logger functions and of the process annotations delivered in this PR (hence no documentation on this).

@kris-nova: the `logger.FInfo` example makes total sense, but if you look at the overall architecture of the `logger` package, you'll see that the implementations decide which writer to choose based on the state of package variables (unless you're in testing mode). IMO, if we declare `io.Writer` parameters in the function signatures, we'll have to rethink this behavior.

As it's submitted here, we don't break with the logger API and we didn't have to change the design. I'd vote for accepting this PR to include annotations for long-running tasks and then reconsider the logger design for the next milestone. Thoughts?

kris-nova: @metmajer at this point I think refactoring the logger makes perfect sense.. it started off one way and and as the project grows it's turning into something else...

My proposal here would be the following:

1. Merge this *as is* to keep our velocity up
2. Open up issues to track the new changes to the logger
3. Update the logger over time

Also another thought would be - would it make sense to move the logger into it's own repository? We own the `kubicorn` org but aren't really using it... thoughts?

metmajer: I'll open up an issue for this. Let's use it to collect some ideas. I don't really feel the need for having the logger moved into its own repository really. While the logger is a central piece, as @prateekgogia rightfully pointed out, we don't necessarily need to invent our own solution. I'll share something in the issue right away. According to the rules, seems like we need 1 more LGTM here :-)

kris-nova:
# tldr 

 - We would like to merge this as is (+1 from me) 
 - @metmajer is going to open a new issue for a broader logger refactor to help get rid of the package state (see comments above) 
 - We need a LGTM from another maintainer
 - We would like feedback on the new issue of how we should solve the logger problem in general as we would like to refactor

Thoughts?

prateekgogia:
If the `golint` has no warnings on this fix, I am ok to merge.

metmajer:
Fixed, ready to merge!

Makefile
```diff
@@ -73,7 +73,7 @@ shell: ## Exec into a container with the kubicorn source mounted inside
 	--rm ${SHELL_IMAGE} /bin/bash
 
 lint: install-tools ## check for style mistakes all Go files using golint
-	golint $(PKGS)
+	~/go/bin/golint $(PKGS)
```
prateekgogia: We shouldn't do this, not everyone has go tools in this location, general approach is to add $GOPATH/bin to $PATH env variable. So, that golint or goimports commands can work. This commands will for sure fail on my host.

metmajer: Argh. This shouldn't have happened. Sorry for this mess, been in a hurry to get the family into vacation and this PR merged 👍

metmajer:
Sorry for this fix mess. Will take care to properly check `golint` and `gofmt` from now on 👍

prateekgogia:
No Worries, thanks for all the help :)

PR #206, opened 2017-08-08
Use travis jobs

aledbf:
With this change this is the output in travis
screenshot from 2017-08-08 11 07 44


kris-nova:
LGTM thanks

PR #207, opened 2017-08-08
makefile: bump version

### LINKED ISSUE #69: We need a version bump make target
Can we add 3 targets to the makefile?

1. Major version bump
2. Minor version bump
3. Patch version bump

They should all edit the `/VERSION` file

xmudrii:
Closes #69 

I found this elegant solution that doesn't seem bad and it works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,15 @@ lint:
 	which golint > /dev/null || go get -u github.com/golang/lint/golint
 	golint $(PKGS)
 
+# versioning
+bump-major:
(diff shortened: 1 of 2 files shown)


kris-nova:
@xmudrii you can use the `Closes #123` syntax and it will actually close the issue btw

kris-nova:
LGTM thanks

xmudrii:
@kris-nova thanks. hmm, I thought `Fixes #123` does the same job, but now I don't remember what I wrote. :D

PR #205, opened 2017-08-08
Lint code using gofmt task

aledbf:

PR #210, opened 2017-08-08
Lint code and fix imports order using gofmt task

aledbf:
replaces #205


kris-nova:
Do we have something in `.travs.yml` to check for this?

aledbf:
@kris-nova no but that should be easy to add. Just abort ci when code with errors is being pushed?

kris-nova:
yep - just have CI yell at us if something fails whatever check you are running here

aledbf:
@kris-nova can I do that in a different PR? This PR is just the result of running `make gofmt` (it changes the files).
I a different PR I'll add a different task to check the code and abort in case of errors

kris-nova:
Sounds good - merging

PR #218, opened 2017-08-08
Added some simple tests

MyDigitalLife:
Just some few simple tests for me to get a feel for how test work in golang. Maybe simple but having them doesn't hurt I think.


kris-nova:
The more tests we have exercising our code, the stronger our code will be.. we just need to make sure we are only testing the bits we plan to keep for a long time ;)

kris-nova:
Please run the headers script so CI will pass :)

PR #223, opened 2017-08-09
ifconfig is deprecated and should be avoided

mzamot:
Some Linux distributions have deprecated the use of ifconfig and route in favor of the software suite iproute2. The command ip should be used instead of ifconfig.


kris-nova:
Excellent point! Can we fix headers so CI passes and we can merge?

mzamot:
Sorry for the ignorance, but exactly what's not passing? :(

kris-nova:
There is a conflict that needs resolved.. and the CI job failed:

You can click on `details` next to the travis icon and see that it failed compile because of the damn bootstrap script again.. I think a rebase should do it.. and we are working on #21 and #132 to solve this issue moving forward

screen shot 2017-08-08 at 20 07 15

kris-nova:
Perfect thank you so much

PR #224, opened 2017-08-09
Add check stages in travis-ci

aledbf:
Example of the validation:
screenshot from 2017-08-08 21 30 56


kris-nova:
The build passed but wasnt reporting - the change looks good - let's see how it does with the rest of the PRs!

Issue #214, opened 2017-08-08
API Validation


kris-nova:
Right now we (dangerously) accept every API that comes our way... can we start to build out API validation using this system:  https://github.com/kris-nova/kubicorn/blob/master/cutil/initapi/init.go

We need a long set of rules, with each rule wrapped in a function (similar to unit tests) and if the function fails we need to have a meaningful message, as well as the name of the function that failed.

The first thing we can check is that each server pool has a value greater than 0

kris-nova:
We will also need tests to prove our validation works and doesn't work (happy and sad paths)

So for the tests, we will need a lot of different profiles, some good, some bad.. and check that good ones work and the bad ones don't work

kris-nova:
We will also need documentation on adding new validation rules, and how every validation rule needs a happy/sad test

kris-nova:
Once we have 1 working validation rule, documentation, and happy/sad tests for the new rule we can close this issue

kris-nova:
Now we have validation #226 and docs #245 so closing!

PR #226, opened 2017-08-09
Add API validation for server pool counts

boingram:
Added checks for non-zero server pool size and non-zero min/max counts for server pools according to #214 

Docs to follow in subsequent PR


kris-nova:
Can we fix the headers? I think you just need to run `make update-headers` and send another commit - then this should be good to merge.

cutil/initapi/validate.go
```diff
@@ -14,4 +14,22 @@
 
 package initapi
 
-// keep
+import (
+	"fmt"
+	"github.com/kris-nova/kubicorn/apis/cluster"
+)
+
+func serverPoolCounts(initCluster *cluster.Cluster) error {
```
kris-nova: Can we break these up into different functions and also give *all* the functions the prefix of `validate`?

For example we want a function called something like `validateServerPoolGreaterThan1` as well as `validateServerPoolMaxCountGreaterThan1` etc etc...

That way we can send another PR that will load these functions dynamically using reflection and we don't have to include the function name in the list anymore!

cutil/initapi/validate_test.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	err := serverPoolCounts(c)
+	if err != nil {
+		t.Fatalf("error message incorrect for valid server pool counts"+
+			"should be: nil\n"+
+			"got:       %v\n", err)
+	}
+}
+
+func TestServerPoolCountsSad(t *testing.T) {
```
kris-nova: Following the syntax above, can we have matching test functions for each of the validate functions and not lump them together? I know it seems like a lot of work, but once we have the validation system in place, the function name will serve as the message that tells us what went wrong 😉

cutil/initapi/validate.go
boingram: Certainly! How exactly would we use reflection to accomplish that? It all seems like dark magic to me, so a real-world example would be pretty helpful.

cutil/initapi/validate_test.go
boingram: Yep, changes on the way!

boingram:
Updated to address comments. 

I was unable to get the make update-headers target to work for me due to sed complaining about an invalid command code, but I inserted the header manually, ran the check-headers script, and that passed

kris-nova:
Gorgeous! Let's merge and we need to open an issue regarding the header bug you experienced. Do you have the error message?

boingram:
Yep, thank god for having way too many iterm tabs still open

    $ make update-headers
    sed: 1: "./cutil/initapi/validat ...": invalid command code .

boingram:
#228 created in reference to the header update script bug mentioned above

PR #236, opened 2017-08-09
Fixed docker build in other directories.

MyDigitalLife:
ATM the docker build script only worked in the docker folder. Now you can just call it from the root like `./docker/build.sh`.


kris-nova:
LGTM Thanks

PR #217, opened 2017-08-08
do: OpenVPN implementation

xmudrii:
Closes #178 
Closes #213 
Closes #171 

Fully implemented OpenVPN solution for DigitalOcean Private Networking.
_There's one TODO, which we will leave for some other PR_

PR #240, opened 2017-08-09
Profile for CentOS - Digital Ocean

mzamot:
Initial creation of bootstrap scripts for CentOS and sample profile.
Still needs inclusion in cmd/create.go


xmudrii:
‎Hi, thank you very much for your contribution.

Looking from the mobile, this looks good. The only thing missing is VPN. Do you plan on adding it, do you need any help? (you can take a look at #217)

If not, you will have to change private IP gathering from `tun0` to `eth1` (or to use Metadata API for Private Networking IP).‎

mzamot:
Got it. I was basing this bootstrap from the simple example. 
Will use your bootstrap as a base and push. Better to use OpenVPN than not.

xmudrii:
That sounds good. If you need any help, I'll be glad to help you. If you need more resources on OpenVPN, take a look at #178.

The only thing it's missing is that one certificate is used across all nodes. I have plan how to fix it, but we agreed to keep it simple for this PR and to fix it in new one.

kris-nova:
Yes let's add the VPN using #217 as an example please

kris-nova:
Can we also add an integration test for this please? It can be a separate PR, just need an issue in place

kris-nova:
Can we please add this to `cmd/create.go` please? We will also need to start working on a piece of documentation that is a large table like `/docs/envvar.md` that will serve as the directory of all of the cloud/operating system permutations we get with kubicorn

mzamot:
Should we keep different scripts per distro for openvpn (openvpnMaster.sh and openvpnNode.sh) or adapt the existing one to support multiple distros?

kris-nova:
The bootstrap scripts are *a bit* undecided at this point.. for now can we please go with the 
> keep it simple
philosophy? 

Do whatever makes the most sense for you, if it means another script, please code it in. I really want to encourage this.

profiles/centosdigitalocean.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+		ServerPools: []*cluster.ServerPool{
+			{
+				Type:     cluster.ServerPoolTypeMaster,
+				Name:     fmt.Sprintf("%s-master", name),
+				MaxCount: 1,
+				Image:    "centos-7-x64",
+				Size:     "1gb",
+				BootstrapScripts: []string{
+					"vpn/meshbirdMaster.sh",
```
kris-nova: Can we use the openVPN scripts now?

profiles/centosdigitalocean.go
```diff
@@ -0,0 +1,66 @@
...
+				},
+			},
+			{
+				Type:     cluster.ServerPoolTypeNode,
+				Name:     fmt.Sprintf("%s-node", name),
+				MaxCount: 1,
+				Image:    "centos-7-x64",
+				Size:     "1gb",
+				BootstrapScripts: []string{
+					"vpn/meshbirdNode.sh",
```
kris-nova: Same as above regarding openVPN

mzamot:
I decided to use separate scripts. Easy to maintain and customize when dealing with multiple distros.

bootstrap/vpn/openvpnNode-centos.sh
```diff
@@ -0,0 +1,24 @@
...
+PRIVATE_IP=$(curl http://169.254.169.254/metadata/v1/interfaces/private/0/ipv4/address)
+
+# OpenVPN
+yum install epel-release -y
+yum install openvpn -y
+
+echo -e ${OPENVPN_CONF} > /etc/openvpn/clients.conf
+
+systemctl start openvpn@clients
+systemctl enable openvpn@clients
```
xmudrii: Add new line here to make sure nothing breaks when we concat two scripts.

bootstrap/vpn/openvpnMaster-centos.sh
```diff
@@ -0,0 +1,122 @@
...
+    <(echo -e '<ca>') \
+    ${KEY_DIR}/ca.crt \
+    <(echo -e '</ca>\n<cert>') \
+    ${KEY_DIR}/clients.crt \
+    <(echo -e '</cert>\n<key>') \
+    ${KEY_DIR}/clients.key \
+    <(echo -e '</key>\n<tls-auth>') \
+    ${KEY_DIR}/ta.key \
+    <(echo -e '</tls-auth>') \
+    > ${OUTPUT_DIR}/clients.conf
```
xmudrii: Add new line here to make sure nothing breaks when we concat two scripts.

mzamot: Perfect, should be fixed.

kris-nova:
LGTM - excellent work here. Thank you so much :rainbow_flag: :heart:

PR #242, opened 2017-08-09
Makefile: Green text instead of black background

roberthbailey:
Instead of setting a black background on the make commands (making them impossible to read without highlighting them), set the text color to be green.


kris-nova:
Welcome to the project!

Issue #42, opened 2017-07-16
We need a file parser


kris-nova:
We need a package/tool/etc that takes a string

`myString` *string* 

and attempts to turn it into another string

`myString` *string*

with the caveat being that we need the first string to represent **some** type of file resource, and the second string to represent it's content.

---

#### Examples:

`~/data` -> `content`
`/tmp/data` -> `content`
`http://nivenly.com/data` -> `content`
`https://nivenly.com/data` -> `content`
`s3://something/data` -> `content`

mariomazo:
@kris-nova could you provide a sample of the ``data`` file. Is do we need some sort of parsing (remove \n and so on)

kris-nova:
@mariomazo we should not alter the file in anyway, only read it to a variable. 

We just need to be able to give a function a string, and it either return a byte slice `[]byte` or an error `error`.

The accepted strings are listed above, but they could be something like the following:
`~/data` should read from `$HOME/data`
`/tmp/data` should read from `/tmp/data`
`http://nivenly.com/data` should send a GET HTTP request to `http://nivenly.com/data` and return the content
`https://nivenly.com/data` should send a GET HTTPs request to `https://nivenly.com/data` and return the content
`s3://something/data` should send a GET HTTPs request to the Amazon S3 endpoint `s3://something/data`

All we are offering is an ability for operators to define friendly strings, and get content from arbitrary locations. Does this make sense?

mariomazo:
I will do it!

PR #252, opened 2017-08-10
simple file parser/fetcher

mariomazo:
implements a simple file parser as required by #42 

if merged please create new issue to add S3 support

### DIFF
--- a/cutil/parser/examples_test.go
+++ b/cutil/parser/examples_test.go
@@ -1,0 +1,27 @@
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package fileresource
+
+import "fmt"
+
+func ExampleReadFromResource() {
+	file, err := ReadFromResource("http://raw.githubusercontent.com/mariomazo/test-files/master/test")
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	fmt.Printf("%v", string(file))
+	//Output:
+	// test
+}
--- a/cutil/parser/fileresource.go
+++ b/cutil/parser/fileresource.go
@@ -1,0 +1,36 @@
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fileresource
+
+import (
+	"net/url"
+	"strings"
+)
+
+// ReadFromResource reads a file from different sources
+// at the moment suppoted resources are http, http, local file system(POSIX)
+func ReadFromResource(r string) ([]byte, error) {
+	switch {
+	case strings.HasPrefix(strings.ToLower(r), "http://") || strings.HasPrefix(strings.ToLower(r), "https://"):
+		url, err := url.ParseRequestURI(r)
+		if err != nil {
+			return []byte(""), err
+		}
+		return readFromHTTP(url)
+
+	default:
+		return readFromFS(r)
+	}
+}
--- a/cutil/parser/fileresource_test.go
+++ b/cutil/parser/fileresource_test.go
@@ -1,0 +1,106 @@
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package fileresource
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"reflect"
+	"testing"
+)
+
+func TestReadFromResource(t *testing.T) {
+	//create local files
+	err := os.Mkdir("./tomeDir", 0755)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.RemoveAll("./tomeDir")
+
+	fileContent := []byte("testing")
+
+	// relative path file
+	tmpfile, err := ioutil.TempFile("./tomeDir", "testfile")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.Remove(tmpfile.Name())
+
+	if _, err := tmpfile.Write(fileContent); err != nil {
+		log.Fatal(err)
+	}
+	if err := tmpfile.Close(); err != nil {
+		log.Fatal(err)
+	}
+
+	// absolute path file
+	tmpfile2, err := ioutil.TempFile(os.TempDir(), "testfile")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.Remove(tmpfile2.Name())
+
+	if _, err := tmpfile2.Write(fileContent); err != nil {
+		log.Fatal(err)
+	}
+	if err := tmpfile2.Close(); err != nil {
+		log.Fatal(err)
+	}
+
+	testURL := "://raw.githubusercontent.com/mariomazo/test-files/master/test"
+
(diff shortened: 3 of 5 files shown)


mariomazo:
forgot to mention,
implements:
- local (POSIX)
- http/https

cutil/parser/fsfileresource/localfs.go
```diff
@@ -0,0 +1,64 @@
...
+// right now just helps us for type enforcment
+type LocalResource struct{}
+
+//NewFileLocalResource creates a new file resource fetcher
+func NewFileLocalResource(fr LocalResource) *LocalResource {
+	return &fr
+}
+
+// Read reads file from a local path and returns as []byte
+func (fr *LocalResource) Read(sourcePath string) ([]byte, error) {
```
prateekgogia: Why is relative path not allowed, I shoud just be able to pass in sth like `-f filename.file` and not give the complete path.

cutil/parser/fsfileresource/localfs.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Here we check if the path starts with ~ or /
+	match, err := regexp.MatchString("[/|~]", sourcePath[:1])
```
prateekgogia: Also, this will crash here is `sourcePath ` is empty. I would suggest to allow relative path, they are easier to add in cmd line.

cutil/parser/fsfileresource/localfs.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if !match {
+		return []byte(""), fmt.Errorf("malformed sourcePath path expected first character to be ~ or /, but got %s", sourcePath)
+	}
+
+	// If absolutePath starts with ~ we search for $HOME
+	// and preppend it to the absolutePath overwriting the first character
+	// TODO: Add Windows support
+	if strings.HasPrefix(sourcePath, "~") {
```
prateekgogia: prefer `filepath.HasPrefix`

cutil/parser/fsfileresource/localfs.go
```diff
@@ -0,0 +1,64 @@
...
+		homeDir := os.Getenv("HOME")
+		if homeDir == "" {
+			return []byte(""), fmt.Errorf("Could not find $HOME")
+		}
+		sourcePath = homeDir + sourcePath[1:]
```
prateekgogia: prefer using `filepath.Join`

cutil/parser/fsfileresource/localfs.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	fqn := sourcePath
```
prateekgogia: `fqn` is not needed, directly use `sourcePath` variable in `ioutil.Readfile()`

cutil/parser/httpfileresource/httpresource.go
```diff
@@ -0,0 +1,68 @@
...
+func (fr *HTTPResource) Read(sourcePath string) ([]byte, error) {
+	// After checking https://www.ietf.org/rfc/rfc3986.txt turns out
+	// that almost everything is a valid URL
+	targetURL, err := url.ParseRequestURI(sourcePath)
+	if err != nil {
+		return []byte(""), err
+	}
+
+	// Here we check if the url starts with http or https
+	match, err := regexp.MatchString("[http|https]", targetURL.Scheme)
```
prateekgogia: ParseRequestURI will return error if HTTP(s) is missing, we don't need this check here.
Look here - 
https://play.golang.org/p/THOCNhqY-2

prateekgogia:
After reading the complete implementation I have couple of doubts:
- How is reading file from S3 any different from HTTP read via Github in your example.
- Do we expect user to provide the source type and source in cmd line like,
` -f http://xyz.com -t http   ||  -f /somepath/file -t fs` ?? 

How I see it is user will provide, something like this 
`-f http://xyz.com || -f s3://abc.aws.com/file | -f /filepath/file`

We can have something generic implemented, when someone provides an input with -f flag to read from. Each resource implements an io.reader and higher level packages can call a function like -  `NewHTTPResource(inputFromUser)  || NewFileResource(inputFromUser)` , this returns an io.Reader and we pass this reader to a generic function which reads stuff from a reader and returns the content
`ReadFromResource(r io.Reader) (string, error) 
`

kris-nova:
I agree with @prateekgogia's concern here, I think the implementations are great we just a way to be lazy about detecting the type here. I am thinking a `type switch` might be the trick we are looking for, or even writing an algorithm to parse the type from the string. Basically we need to be able to say "Here is my string, give me some data, I don't care where it comes from"

kris-nova:
The impls are super beautiful though - nice work

cutil/parser/fsfileresource/localfs_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"reflect"
+	"testing"
+)
+
+func TestLocalResource_Read(t *testing.T) {
```
MyDigitalLife: I think there should also be some tests to see if throwing errors works a intended. So tests of resources that don't exists. ATM it looks like you only test with existing files if i'm not mistaking. This goes the same for the http resources tests.

cutil/parser/fsfileresource/localfs.go
mariomazo: it wasn't in any of the example so I went "monkey sees monkey does"

mariomazo:
Hello @prateekgogia @kris-nova @MyDigitalLife

First thanks for the feedback I will correct the missing testing and ``filepath``

Then I will have questions regarding the implementation. Initially I had something like
```
type RemoteResource struct{}
```

but then I thought that for http might start basic, but then maybe we want basic auth, certificates, and whoknkows, then trow in there s3 and whatever comes in the future
```
type RemoteResource struct {
    basic_auth
    more_stuff
    *aws_session
}
```
This can get crazy! So that was my reasoning to split them

if you think thats the way to go we can still go for it and have a switch like
that was my initial approach
```
 switch scheme {
case http(s)
   readHTTP()
case s3
   readS3()
}
```
that unfortunately would not solve the @kris-nova desire to have only one read. In my mind having local and remote still makes sense as they are quite different and can grow in different directions, but definitely could be something like

```
if valid requestURI
 do the switch scheme (http(s)/s3)
default:
 // is invalid for a request
 is it a valid path?
 do readLocalFS
end
```

any tips in how to validate if a thing is valid path are appreciated it. or should I just stat() it?

mariomazo:
umm probably a more elegant consistent version would be ``ReadAll(r io.Reader) ([]byte, error)``

cutil/parser/httpfileresource/httpresource.go
```diff
@@ -0,0 +1,68 @@
...
+	"net/url"
+	"regexp"
+)
+
+// HTTPResource could hold explicit information for HTTP calls
+// right now just helps us for type enforcment
+type HTTPResource struct{}
+
+// NewHTTPResource creates a new file resource fetcher
+func NewHTTPResource(fr HTTPResource) *HTTPResource {
```
kris-nova: I don't understand why we are taking in a struct (that has no members) and returning a pointer to that struct?

`new` to me usually means initialize a struct with arbitrary inputs and return a point to a *newly* initialized struct - if it is initialized before we call new... then it isn't really new?

kris-nova:
@mariomazo 

> any tips in how to validate if a thing is valid path are appreciated it. or should I just stat() it?

My original thought was that we would have a small system that tries to reason about a path based on what's in it..

for instance..

```go
path := "https://something.com/path

// ------- small system -------
if strings.HasPrefix(path, "https://") {
    fr := hr.NewHTTPResource() // returns a FileResource interface
    bytes, err := fr.Read(path)
    if err != nil {
         // return err
    }
    // return bytes
} else if strings.HasPrefix(path, "http://") {
    // Same logic as above
} else if strings.HasPrefix(path, "s3://" {
    // Same logic as above
} else if strings.HasPrefix(path, "/" {
    fr := hr. NewFileLocalResource() // returns a FileResource interface
    bytes, err := fr.Read(path)
    if err != nil {
         // return err
    }
    // return bytes
} else if strings.HasPrefix(path, ".") {
    // Same logic as above
} else if strings.HasPrefix(path, "~" {
   // Same logic as above, but also call Expand()
} 
// etc..
```

We could (and probably should) run this in a switch but I don't care how we do it.. just want it to be easy to understand for others.

Does this make sense?

cutil/parser/httpfileresource/httpresource.go
mariomazo: Well, my idea was that those empty structs where just place holders, so if in the future something was put in there, the usage would not change that much.

kris-nova:
@mariomazo 

> finally should we go for ReadFromResource(r io.Reader) (string, error) or ReadFromResource(r string) ([]byte, error) ?

I vote  `ReadFromResource(r string) ([]byte, error)`

mariomazo:
Thats exactly what I proposed in my "no so last comment", and if you go for ``ReadFromResource(r string) ([]byte, error)`` it also looks fine to me, so I can move one with the re-implemintation tonight

cutil/parser/fsfileresource/localfs.go
mariomazo: @prateekgogia "filepath.HasPrefix exists for historical compatibility and should not be used."
```go
// HasPrefix exists for historical compatibility and should not be used.
func HasPrefix(p, prefix string) bool {
    return strings.HasPrefix(p, prefix)
}
```

prateekgogia: Good to know, thanks :)

prateekgogia:
@mariomazo Thanks this looks a lot simplified now, though there can be some more improvements. we can do them later, like we can just return string. Right now we are returning bytes slice and most of the places we are constructing the slice, we can avoid doing that by simply returning strings. Its fine for now as I said, it can be an improvement later.

prateekgogia:
LGTM

PR #265, opened 2017-08-13
cmd: use argument for name

### LINKED ISSUE #260: Cluster name as implicit argument
So right now we use a cluster name to identify clusters.. but the user has to pass `-n` or `--name` in order to define the cluster such as:

```
kubicorn apply -n mycluster
```

Can we have an implicit `-n` for the first argument after the command so that the user can use simply type:

```
kubicorn apply mycluster
``` 

This also frees up the `-n` flag for the `kubicorn snap` command to be used for `-n` `--namespaces` which makes much more sense to a `kubectl` user.

xmudrii:
Closes #260 

Auto completions are temporarily broken (and disabled), but I'll fix it in another PR.


cmd/apply.go
```diff
@@ -29,15 +29,27 @@ import (
...
-	Use:   "apply [-n|--name NAME]",
+	Use:   "apply <NAME>",
 	Short: "Apply a cluster resource to a cloud",
 	Long: `Use this command to apply an API model in a cloud.
 
 This command will attempt to find an API model in a defined state store, and then apply any changes needed directly to a cloud.
 The apply will run once, and ultimately time out if something goes wrong.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) != 1 {
+			ao.Name = strEnvDef("KUBICORN_NAME", "")
```
kris-nova: Instead of defaulting to `""` should we use the namer to generate a random name?

cmd/apply.go
```diff
@@ -29,15 +29,27 @@ import (
...
+		} else {
```
kris-nova: Can we check for `> 1` as well and throw an error that too many arguments were passed?

cmd/apply.go
xmudrii: Does we have a namer somewhere or I need to code it?

kris-nova:
We will need to update our docs as well since we are removing `-n`

xmudrii:
@kris-nova fixed docs as well :)

kris-nova:
Great - I will update my blog and merge

PR #267, opened 2017-08-13
[WIP] signals: implement signals and retry packages

### LINKED ISSUE #44: Signal Handler Interface
We need to create a global signal handler, that will reference an interface. Each reconciler should implement said interface so that we can handle POSIX signals globally.

If you are interested in coding we should talk, because I would want to hear your ideas! Otherwise I might just code this.

### LINKED ISSUE #203: Implement retry package and new signal handler
I have some ideas for a POSIX compliant signal handler, and a clever re-try package that we can use throughout the program that will respect the new signal handler, as well as make it easy to retry functions for an arbitrary amount of time.

xmudrii:
Closes #44 
Closes #203 

Still WIP


cutil/signals/signals.go
```diff
@@ -0,0 +1,50 @@
...
+
+	go handle(signals)
+}
+
+func handle(signals <-chan os.Signal) {
+	for {			
+			select {
+				case s := <-signals:
+					switch {
+						case s == os.Interrupt:
```
kris-nova: Should we support more signals here? What about `SIGQUIT` for instance..

cutil/signals/signals.go
```diff
@@ -0,0 +1,50 @@
...
+// limitations under the License.
+
+package signals
+
+import (
+	"os/signal"
+	"os"
+)
+
+var State os.Signal
```
kris-nova: Can we unexport this? We absolutely never want an outside package to be able to change this.

kris-nova: We can have an exported function that reads it - so other packages can get the state, but we just don't want it to be mutated by anything other than this package.

cutil/signals/signals.go
```diff
@@ -0,0 +1,50 @@
...
+	for {			
+			select {
+				case s := <-signals:
+					switch {
+						case s == os.Interrupt:
+							if State == nil {
+								State = os.Interrupt
+								continue
+							}
+							os.Exit(1)
```
kris-nova: We will need to start documenting exit codes in `/docs` as a table

cloud/digitalocean/reconciler.go
```diff
@@ -110,17 +110,16 @@ var createdResources = make(map[int]cloud.Resource)
 func (r *Reconciler) Reconcile(actualCluster, expectedCluster *cluster.Cluster) (*cluster.Cluster, error) {
 	newCluster := newClusterDefaults(r.Known)
 
+	h := signals.NewSignalHandler(10 * time.Minute)
```
kris-nova: Can we move this up to `Init` please? It's not really part of the `Reconcile`

cloud/digitalocean/reconciler.go
```diff
@@ -213,10 +212,14 @@ func newClusterDefaults(base *cluster.Cluster) *cluster.Cluster {
 	return new
 }
 
-func handleCtrlC(c chan os.Signal) {
-	sig := <-c
-	if sig == syscall.SIGINT {
-		sigCaught = true
-		logger.Critical("Detected SIGINT. Please be patient while kubicorn cleanly exits. Maybe get a cup of tea?")
-	}
+func handleCtrlC(h *signals.Handler) {
```
kris-nova: Can we give this a name that corresponds to the POSIX constant we are handling?

`handleSigInt`

cloud/digitalocean/reconciler.go
```diff
@@ -213,10 +212,14 @@ func newClusterDefaults(base *cluster.Cluster) *cluster.Cluster {
...
+	go func() {
+		for {
+			if h.GetState() != 0 {
+				sigCaught = true
+				logger.Critical("Detected signal. Please be patient while kubicorn cleanly exits. Maybe get a cup of tea?")
```
kris-nova: Can we get a `sync.Mutex` on this log line please? I notice it will pop up a lot..

cutil/retry/retry.go
```diff
@@ -0,0 +1,54 @@
...
+package retry
+
+import (
+	"os"
+	"time"
+
+	"github.com/kris-nova/kubicorn/cutil/signals"
+)
+
+type Retry interface {
```
kris-nova: So I think we need an interface.. but I am wondering if the `Worker` is what needs it.

See my note below

cutil/retry/retry.go
```diff
@@ -0,0 +1,54 @@
...
+	Work()
+}
+
+type Worker struct {
+	Function func()
```
kris-nova: So this limits our `Worker` to only being able to call a function.. if we could implement an interface here instead of `Retry` we could define our own implementation.. and then the `Worker` struct could just call `interface.Run()` for instance.. This is probably the worst way of explaining it ever.. but if you need me to clarify just ping me

docs/EXIT_CODES.md
```diff
@@ -0,0 +1,10 @@
+# Exit Codes
+
+This file documents exit codes which `kubicorn` will return on errors.
+
+| Exit code | Meaning |
+|---|---|
+| 1 | General error |
```
kris-nova: Can we use the POSIX constant names here as well? `SIGINT` `SIQTERM` etc

kris-nova:
@xmudrii can you fix the content and then we can disable the tests and merge.

I will look at fixing the tests once we have this merged if that is okay!

Issue #270, opened 2017-08-13
bug - ssh agent support has been broken


paulczar:
The following PRs have resulted in the ssh agent support not working for passphrase enabled private keys:

[pr-162](https://github.com/kris-nova/kubicorn/pull/162/files#diff-2e2495d0054dc0c7c5159b1399dde72bR62) - the reordering of some of the bits here made the sshagent section try to fire after ssh asks for password.

[pr-66](https://github.com/kris-nova/kubicorn/pull/66/files#diff-8836282bf16b255b154ed71be22359e5R39) - kicks off a chain that tries to fingerprint the private key and asks for the passphrase.

[pr-162](https://github.com/kris-nova/kubicorn/pull/162/files#diff-2e2495d0054dc0c7c5159b1399dde72bR66) - I suspect this might also cause skip using ssagent, but I haven't seen it hit yet.

kris-nova:
meep yes they have - i noticed SSH was broken for me the other day so I clobbered everything (which was super not okay I am so sorry) 

I noticed you opened #271 that looks completely sane and should fix this.. but I really want us to get some tests that demonstrate all of these things are working as expected.

kris-nova:
I just opened #272 regarding the tests - we can merge #271 and hope for the best until then!

paulczar:
I've seen a number of tools that require the use of ssh-agent for handling keys, I wonder if that might be a more sane way to handle this?  we can be "do whatever i don't care lulz just have the key in your agent" with maybe a fullback to require ( or create in _state a one shot passphraseless keypair ).

currently without agent but with a passphrase it asks what feels like a dozen times for the passphrase, given nearly everybody has a passphrase ( right? really please tell me this is true!! ) on their default key ( .ssh/id_rsa ) it makes sense to me at least to assume they can utilize ssh-agent ( plus also means doing it through a bastion/jump with agent forwarding enabled ) works which reduces the chance some clown would upload their key to a remote machine.

xmudrii:
I've had idea of making sort of in-memory agent that would be present only for a command session. So once command finishes, everything is discarded. That way we don't have to play with security by saving stuff ourself, yet we are not extremely annoying.  

However, I'm not sure is this possible with Go and how it can be done. Maybe I could dig something out, but there is some stuff that most be done for 0.1.00 firstly.

xmudrii:
‎Update: It could be done quite easy. I can give it a try tonight. @kris-nova WDYT?

[...]

PR #271, opened 2017-08-13
Fixes kubeconfig.go ability to use ssh agent

paulczar:
Partially addresses #270 

We should call `getSigner` only if ssh agent is not enabled.


kris-nova:
Not to be a stickler but we have had a few issues with the SSH stuff acting up on us - any way we could write a happy/sad test(s) for with and without an agent? Maybe run it against a container?

kris-nova:
Okay I opened up an issue for the new SSH tests #272 - we don't need it in this PR 

LGTM

PR #273, opened 2017-08-14
New interface proposal 

kris-nova:
CC @prateekgogia @xmudrii @roberthbailey

### DIFF
--- a/docs/reconciler-proposal/kris-1/interface.go
+++ b/docs/reconciler-proposal/kris-1/interface.go
@@ -1,19 +1,62 @@
-package kris_1
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
+package cloud
 
-type Reconciler interface {
-	Init() error
-	GetActual() (*cluster.Cluster, error)
-	GetExpected() (*cluster.Cluster, error)
-	Reconcile(actualCluster, expectedCluster *cluster.Cluster) (*cluster.Cluster, error)
-	Destroy() (*cluster.Cluster, error)
+import (
+	"github.com/kris-nova/kubicorn/apis/cluster"
+)
+
+// Model will model resources to an API representation of infrastructure
+type Model interface {
+
+	// Get will return the map of ordered resources against a cloud
+	Get() map[int]Resource
 }
 
+// Reconciler will create and destroy infrastructure based on an intended state. A Reconciler will
+// also audit the expected and actual state.
+type Reconciler interface {
+
+	//SetModel will set the model to use for the cloud reconciler
+	SetModel(model Model)
+
+	// GetActual will audit a cloud and return the API representation of the current resources in the cloud
+	GetActual(known *cluster.Cluster) (*cluster.Cluster, error)
+
+	// GetExpected will audit a state store and return the API representation of the intended resources in the cloud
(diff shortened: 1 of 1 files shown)


docs/reconciler-proposal/kris-1/interface.go
```diff
@@ -0,0 +1,62 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cloud
+
+import (
```
roberthbailey: nit: if you only have a single import you can put it onto one line

docs/reconciler-proposal/kris-1/interface.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/kris-nova/kubicorn/apis/cluster"
+)
+
+// Model will model resources to an API representation of infrastructure
```
roberthbailey: nit: Comments should end with a period. 

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

(also many places below)

docs/reconciler-proposal/kris-1/interface.go
```diff
@@ -0,0 +1,62 @@
...
+type Model interface {
+
+	// Get will return the map of ordered resources against a cloud
+	Get() map[int]Resource
```
roberthbailey: Why a map if they are ordered? Why not a slice? Do we expect iteration across the objects to pretty much always happen in order?

docs/reconciler-proposal/kris-1/interface.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// Reconciler will create and destroy infrastructure based on an intended state. A Reconciler will
+// also audit the expected and actual state.
+type Reconciler interface {
+
+	//SetModel will set the model to use for the cloud reconciler
```
roberthbailey: nit: space between // and start of comment

docs/reconciler-proposal/kris-1/interface.go
```diff
@@ -0,0 +1,62 @@
...
+	// GetActual will audit a cloud and return the API representation of the current resources in the cloud
+	GetActual(known *cluster.Cluster) (*cluster.Cluster, error)
+
+	// GetExpected will audit a state store and return the API representation of the intended resources in the cloud
+	GetExpected(known *cluster.Cluster) (*cluster.Cluster, error)
+
+	// Reconcile will take an actual and expected API representation and attempt to ensure the intended state
+	Reconcile(actual, expected *cluster.Cluster) (*cluster.Cluster, error)
+
+	// Destroy will take an actual API representation and destroy the resources in the cloud
```
roberthbailey: Why does destroy need an actual? Will it only remove resources found in the enclosed representation? Or will it try to determine any skew based on the cluster definition and clean up everything it can find?

docs/reconciler-proposal/kris-1/interface.go
kris-nova: Order is critical here - and building a hash map is one way of preserving order... we go `0:N` for a create and `N:0` for a delete. 

By using a map instead of a slice, we can index them and use the indexes in an arbitrary way to point to resources down the line as seen here https://github.com/kris-nova/kubicorn/blob/master/cloud/amazon/model.go#L43

docs/reconciler-proposal/kris-1/interface.go
kris-nova: This is a great point.

With this approach `Destroy` would take an `actual` which is what is running in the cloud and destroy the resource that are modeled out of the API. I don't see a case where we would ever have a skew unless the calling code passed in a representation that was different than what an `actual` would be at runtime. 

In other words, `Destroy` takes an actual so we know we are deleting resources that actually exist. If a user passed in an `expected` API (or any API for that matter) they would inherently be using `Destroy` wrong.

It would be possible to do an asserting at runtime during the beginning of a `Destroy` but that would double our efforts of querying the cloud as we would expect a user to call `GetActual` right before we call the same method to assert they are the same.

What are your thoughts here? Does this seem sane?

kris-nova:
@roberthbailey going to merge and get your nit's in the follow up PR I am working on!

PR #274, opened 2017-08-14
Minor log message change - removeing "subnet"

matyix:
Minor message change - removed subnet from the log message - and a question.

I have found this while investigating a `potential ???` issue and will track this one below in a different issue and sent a fix in case it's valid - it's just a heads un and very much interested in your opinion. In case a user has changed/regenerated his local keys, and a cluster is created with the new key pair but same name, the old keypairs are still available on the AWS side. The cluster is created (with the same name - in my case by a CI system) but the end user/test suite wont be able to SSH into it. Obliviously this is up to interpretation and not a likely scenario - just wandering whether it makes sense to check or log whether the user's key has changed and the public fingerprints are different. 

Thank you.

### DIFF
--- a/cloud/amazon/resources/keypair.go
+++ b/cloud/amazon/resources/keypair.go
@@ -80,7 +80,7 @@ func (r *KeyPair) Actual(known *cluster.Cluster) (cloud.Resource, error) {
 func (r *KeyPair) Expected(known *cluster.Cluster) (cloud.Resource, error) {
 	logger.Debug("keypair.Expected")
 	if r.CachedExpected != nil {
-		logger.Debug("Using keypair subnet [expected]")
+		logger.Debug("Using keypair [expected]")
 		return r.CachedExpected, nil
 	}
 	expected := &KeyPair{


kris-nova:
I think it would be valuable to check that the fingerprints are as expected. I would almost expect `kubicorn` to error if that was the case as a user shouldn't be naming the keys the same anyway.

Another (more elegant) approach to solving this would be to take a hash of the public key fingerprint, and use that for the name. That was a user can flip keys as much as they want and `kubicorn` will seamlessly handle them well.

kris-nova:
In fact I like that idea so much I just opened up #279 :smile:

kris-nova:
This change looks great - thanks for contributing to kubicorn!

mariomazo:
@kris-nova Maybe I misunderstood, but as I see it ,and my experience with hashicrop packer
is that at the end of the day you end with 200 resources, in this case 200 keypairs named as a hash. Doing the cleanup is really difficult and potentially dangerous.

Issue #183, opened 2017-08-07
Write the cluster API to a file instead of using inject


kris-nova:
So write now we are doing some hacky shit...

We set one-off key/value pairs in Go.. and then hack them into the bash script.. 
https://github.com/kris-nova/kubicorn/blob/master/cloud/digitalocean/resources/droplet.go#L178-L179

### The problem

We *dont* want to encourage a `text/template` pattern here, so we inject one off values
We need data from the API at runtime in these scripts

### The new solution

Can we write a `json` marshalled representation of the API (at runtime) to `/etc/kubicorn/api.conf` on each of the nodes we create? 

We can do this instead of doing the `inject` logic we are doing now..

This way the bash scripts can use `jq` to get any/all data they could need.. this creates a rich user experience and will help users who wish to bring their own bash bootstrap scripts

kris-nova:
We can get rid of this crap: https://github.com/kris-nova/kubicorn/blob/master/bootstrap/digitalocean_k8s_ubuntu_16.04_master.sh#L5-L16

PR #281, opened 2017-08-15
Write cluster as JSON

boingram:
Wanted to get a PR out to review the approach for #183 

I added a JSONFileSystemStore that's heavily influenced by FileSystemStore. Assuming this is the way to go, in subsequent PRs I will update the various commands that write to a state store to also write to a JSON state store as well as updating the bash scripts to look at the JSON file.

### DIFF
--- a/state/filenames.go
+++ b/state/filenames.go
@@ -15,7 +15,7 @@
 package state
 
 const (
-	ClusterFile     = "cluster.yaml"
-	ClusterJsonFile = "api.conf"
+	ClusterJsonFile = "cluster.json"
 	ClusterJsonPath = "/etc/kubicorn/"
+	ClusterYamlFile = "cluster.yaml"
 )
--- a/state/filenames.go
+++ b/state/filenames.go
@@ -16,6 +16,5 @@ package state
 
 const (
 	ClusterJsonFile = "cluster.json"
-	ClusterJsonPath = "/etc/kubicorn/"
 	ClusterYamlFile = "cluster.yaml"
 )
--- a/state/filenames.go
+++ b/state/filenames.go
@@ -16,4 +16,6 @@ package state
 
 const (
 	ClusterFile = "cluster.yaml"
+	ClusterJsonFile = "api.conf"
+	ClusterJsonPath = "/etc/kubicorn/"
 )
--- a/state/filenames.go
+++ b/state/filenames.go
@@ -15,7 +15,7 @@
 package state
 
 const (
-	ClusterFile = "cluster.yaml"
+	ClusterFile     = "cluster.yaml"
 	ClusterJsonFile = "api.conf"
 	ClusterJsonPath = "/etc/kubicorn/"
 )
--- a/state/jsonfs/impl.go
+++ b/state/jsonfs/impl.go
@@ -1,17 +1,31 @@
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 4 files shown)


state/filenames.go
```diff
@@ -15,5 +15,7 @@
 package state
 
 const (
-	ClusterFile = "cluster.yaml"
+	ClusterFile     = "cluster.yaml"
+	ClusterJsonFile = "api.conf"
```
kris-nova: Can we stick with the `cluster.$encoding` format here? So we would use `cluster.json`?
The reason for this is for when we build snapshots (in the future) we are going to want the files named to represent the cluster, and the apps on top.

Also if we are introducing new filenames, can we patch the old constant to use `ClusterYamlFile` to keep the convention throughout the package?

state/jsonfs/impl.go
```diff
@@ -0,0 +1,140 @@
...
+	return os.RemoveAll(fs.AbsolutePath)
+}
+
+func (fs *JSONFileSystemStore) GetCluster() (*cluster.Cluster, error) {
+	cluster := &cluster.Cluster{}
+	configBytes, err := fs.read(state.ClusterJsonFile)
+	if err != nil {
+		return cluster, err
+	}
+	err = yaml.Unmarshal(configBytes, cluster)
```
kris-nova: `yaml.Unmarshal`?

state/jsonfs/impl.go
```diff
@@ -0,0 +1,140 @@
...
+		return []byte(""), err
+	}
+	return bytes, nil
+}
+
+func (fs *JSONFileSystemStore) Commit(c *cluster.Cluster) error {
+	if c == nil {
+		return fmt.Errorf("Nil cluster spec")
+	}
+	bytes, err := yaml.Marshal(c)
```
kris-nova: `yaml.Marshal`?

state/filenames.go
boingram: Certainly!

state/jsonfs/impl.go
boingram: eeeeeeek, I was copy/pasting this file (with json.Marshal) around as I debated over the best home for it. Guess I grabbed fs/impl.go at some point. I'll get it fixed tonight.

state/filenames.go
```diff
@@ -15,5 +15,7 @@
 package state
 
 const (
-	ClusterFile = "cluster.yaml"
+	ClusterJsonFile = "cluster.json"
+	ClusterJsonPath = "/etc/kubicorn/"
```
kris-nova: Sorry I just noticed this.. it looks like we are hard coding all JSON files to `/etc/kubicorn`.. I am wondering if we can't follow suit with the `fs` implementation and have the value here be optionally overwritten at runtime. 

There are command line flags, and environmental variables that can be set that will drop a cluster off here if needed. Seems like a more flexible approach than hard coding.. What are your thoughts?

state/jsonfs/impl.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (fs *JSONFileSystemStore) Exists() bool {
+	if _, err := os.Stat(fs.AbsolutePath); os.IsNotExist(err) {
+		return false
+	}
+	return true
+}
+
+func (fs *JSONFileSystemStore) write(relativePath string, data []byte) error {
+	fqn := fmt.Sprintf("%s/%s", fs.AbsolutePath, relativePath)
```
prateekgogia: `filepath.Join()`

state/jsonfs/impl.go
```diff
@@ -0,0 +1,140 @@
...
+	os.MkdirAll(path.Dir(fqn), 0700)
```
prateekgogia: Check error for `mkdir`

state/jsonfs/impl.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (fs *JSONFileSystemStore) Commit(c *cluster.Cluster) error {
+	if c == nil {
+		return fmt.Errorf("Nil cluster spec")
+	}
+	bytes, err := json.Marshal(c)
+	if err != nil {
+		return err
+	}
+	fs.write(state.ClusterJsonFile, bytes)
```
prateekgogia: Same can we check write error.

state/filenames.go
boingram: It definitely makes more sense, I was going based off of what was in the issue. Since that's more flexible than I realized, I 100% agree.

PR #296, opened 2017-08-16
KUBICORN_FORCE_DELETE_KEY envvar added

### LINKED ISSUE #285: Force remove SSH keys
Right now we hard code `force=false` for deleting keypairs. This means that we will always orphan a keypair in AWS if a user creates a cluster with a unique keypair and then destroys it after.

Can we please expose this flag via an environmental variable, document the environmental variable in `/docs/envvar.md`, and patch this?

Suggested variable name: `KUBICORN_FORCE_DELETE_KEY` 

The system as it stands today in AWS is here: https://github.com/kris-nova/kubicorn/blob/master/cloud/amazon/resources/keypair.go#L137-L151

ellenkorbes:
Closes #285.

### DIFF
--- a/cloud/amazon/resources/keypair.go
+++ b/cloud/amazon/resources/keypair.go
@@ -16,8 +16,8 @@ package resources
 
 import (
 	"fmt"
-	"strings"
 	"os"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/kris-nova/kubicorn/apis/cluster"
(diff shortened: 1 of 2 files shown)


cloud/amazon/resources/keypair.go
```diff
@@ -134,7 +135,11 @@ func (r *KeyPair) Apply(actual, expected cloud.Resource, applyCluster *cluster.C
 }
 func (r *KeyPair) Delete(actual cloud.Resource, known *cluster.Cluster) (cloud.Resource, error) {
 	logger.Debug("keypair.Delete")
-	force := false
+	if os.Getenv("KUBICORN_FORCE_DELETE_KEY") != "" {
```
prateekgogia: in this case if I run it as `export KUBICORN_FORCE_DELETE_KEY=false`, it will still delete the AWS key.

xmudrii: You will laso need to get `force` out of the block. Such as:
```go
var force bool
if ... {
  force = true
}...
```

Issue #286, opened 2017-08-16
Use unique name (within VPC/region) for a security group


matyix:
SecurityGroup names within one VPC or region should be unique - currently we are using the name of the cluster, which makes deploying a cluster with the same name/or upscaling the cluster impossible due to duplicate name conflicts. From the AWS constraints lists:

```
InvalidGroup.Duplicate

You cannot create a security group with the same name as an existing security group in the same VPC, or the same region (EC2-Classic).
```

For this I am suggesting a time ordered UUID generator (under cutil package) and appending/using  this as the name of the security group. Are you OK with that? The generator looks stg like this - though can use any other UUID generator, should you suggest one - just choose the time ordered one as it helps with listing (when checking the AWS console).

```
// Generates a time ordered UUID. Top 32b timestamp bottom 96b random.
func TimeOrderedUUID() string {
	unixTime := uint32(time.Now().UTC().Unix())
	randomPart1 := rand.Uint32()
	randomPart2 := rand.Uint32()
	randomPart3 := rand.Uint32()
	return fmt.Sprintf("%08x-%04x-%04x-%04x-%04x%08x",
		unixTime,
		uint16(randomPart1>>16),
		uint16(randomPart1&0xffff),
		uint16(randomPart2>>16),
		uint16(randomPart2&0xffff),
		randomPart3)
}
```

kris-nova:
The proposal looks good - I just ask that we combine the cluster name with the UUID such as:

```go
uniqueName := fmt.Sprintf("%s-%s, clusterName, uuid)
```

Are you interested in implementing this change?

PR #300, opened 2017-08-17
Kubicorn #286 - Use unique name (within VPC/region) for a security group

matyix:
This PR is related to #286 - a time ordered UUID generator is added and the unique name is applied when specifying the name of the security group in the default clusters. 

Ideally I wanted to have this included in the `securitygroup.go` - however while working on cluster update found this causes different issues during reconcile and state management. Once the new refactor is finished and also a different cluster update PR is merged we can return to this - for now this leaves the uniqueness generation (as AWS does) to the end user's discretion, and serves as an example in the cluster templates, and providing a UUID generator.

### DIFF
--- a/cutil/uuid/uuid.go
+++ b/cutil/uuid/uuid.go
@@ -1,0 +1,36 @@
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


kris-nova:
LGTM - can you fix the conflict please and we will merge

matyix:
Reverted back to the version in [Making make ci work again](https://github.com/kris-nova/kubicorn/commit/50b8b49847b929d050384df8291eedfa73dacfb4) for the examples to be consistent with all the other providers.

kris-nova:
Thank you so much for fixing the conflict.. sorry to be a nag but the headers files are currently failing in CI

```0.74s$ make check-headers
incorrect license header: ./cutil/uuid/uuid.go
make: *** [check-headers] Error 1
```

LMK if you can take a look .. you should just be able to delete the headers entirely and run `make update-headers` to auto-add one, and you can verify the change with `make check-headers`

Cheers

matyix:
Sure - my bad, haven't read the contrib guidelines.

kris-nova:
Woo! Nice job!

PR #304, opened 2017-08-17
Save state as JSON when running create command

boingram:
Saving state as JSON when running the create command, per #183 

I wanted to check the initial approach before committing more. Whenever we write to state, we do a second write to the JSON state store


boingram:
On a somewhat related note, I originally started out doing this on the apply command. Creating tests for that became fairly complex as it has several dependencies -- SSH, the cluster actually being up and running so that it can do the Kubeconfig stuff, etc. Is the best practice for automated testing that to extract any dependencies into interfaces that I pass into the command, or are we just better off doing integration tests that peek at the state stores?

kris-nova:
@boingram I always prefer starting with integration tests as they offer high confidence early. 

One thing I want to be very clear on is that #183 we are referring to writing a one time copy of the API to disk on each of of the **kubernetes nodes** so that the bootstrap scripts can use them instead of injecting variables into the script. 

I still thinking being able to save as json will be useful, but I think that the real value of saving a `json` representation of an API would come in the form of `kubicorn get $name -ojson` or something similar. 

Does this make sense and should we close this PR? Let me know as I am always happy to learn more.

boingram:
@kris-nova Thanks for the testing advice! I'll close the PR, but before I do that, I'd like to make sure I understand things correctly. Taking AWS as our cloud provider of choice for this example, we'd like to write the cluster state as JSON at some point in time before our bootstrap scripts run so that we don't have to inject anything. Since in AWS, the bootstrap scripts run as user data, we'll need to add another bootstrap script that gets run first in the big user data script that writes the JSONified cluster state to disk. Is that the approach desired here?

kris-nova:
Woops! Just noticed this was waiting on me - will review and get back ASAP 

Meep!

Sorry!

PR #306, opened 2017-08-18
Update DO ssh key documentation

clintkitson:
This PR updates the DO example documentation for the SSH key configuration. Currently the code requires that the DO SSH key name be the same as the cluster name but this is not documented.

The following error results when a SSH key is not known by DO that matches the name of the cluster.
`Error during apply of atomic reconciler, attempting clawback: strconv.Atoi: parsing "": invalid syntax`


xmudrii:
I believe this is a bug and I'll try to see can I fix it around.

xmudrii:
@clintkitson Thank you very much for reporting the issue. I created a PR to fix this behavior, #308. This should do the job.

PR #311, opened 2017-08-18
Added cloudid option

### LINKED ISSUE #291: Refactor clusters for cloud provider ids.
Cloud providers like Google want you to do API calls to a project id. At the moment for the GCE implementation the Cluster name is used to provider this project id to the API. This should be 2 different fields: A `kubicorn cluster name` and a `cloud resource id`. Not all cloud providers will need a cluster resource id so this should be something that is optional but required when choosing providers like GCE.

As discussed in #226 in [this](https://github.com/kris-nova/kubicorn/pull/266/files/ccc7dfe9bfa7250b29a9b29aa22ed7653d82f341#r133355582) comment.

MyDigitalLife:
For GCE atm you needed to have the kubicorn name and the project-id be the same. This PR adds a parameter cloudid to the create command to be able to set the project-id differently then the kubicorn name. It also requires the cloudid parameter but only for GCE.

Closes #291

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -39,7 +39,7 @@ type CreateOptions struct {
 var co = &CreateOptions{}
 
 var createCmd = &cobra.Command{
-	Use:   "create [NAME] [-p|--profile PROFILENAME] [-ci|--cloudid CLOUDID]",
+	Use:   "create [NAME] [-p|--profile PROFILENAME] [-c|--cloudid CLOUDID]",
 	Short: "Create a Kubicorn API model from a profile",
 	Long: `Use this command to create a Kubicorn API model in a defined state store.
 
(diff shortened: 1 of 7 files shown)


cmd/create.go
```diff
@@ -127,13 +129,17 @@ func RunCreate(options *CreateOptions) error {
...
+	var newCluster *cluster.Cluster
 	if _, ok := profileMapIndexed[options.Profile]; ok {
-		cluster = profileMapIndexed[options.Profile].profileFunc(name)
+		newCluster = profileMapIndexed[options.Profile].profileFunc(name)
 	} else {
 		return fmt.Errorf("Invalid profile [%s]", options.Profile)
 	}
 
+	if newCluster.Cloud == cluster.CloudGoogle && options.CloudId == "" {
+		return fmt.Errorf("--cloudid is required for google cloud.")
```
kris-nova: So we don't want to assume that user is calling from the command line, as this will be ran as a controller in a pod in the future. Can we please update this to suggest the name of the field, and not a command line flag.

```go
    return fmt.Errorf("CloudID is required for Google Cloud")
```

PR #313, opened 2017-08-19
Edit command

### LINKED ISSUE #290: kubicorn edit command
As defined in the `ark` integration here https://github.com/kris-nova/kubicorn/issues/195#issuecomment-322863670 we will need a way to easily edit a cluster configuration in a local text editor. The workflow of the process should look like:

## kubicorn edit
a. Read the snapshot from the state store.
b. Write the snapshot to the local fs as a tmp file
c. Open the file in user's text editor
d. Save the tmp file
e. Write to state store

MyDigitalLife:
Closes #290


cmd/edit.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+var eo = &EditOptions{}
+
+var editCmd = &cobra.Command{
+	Use:   "edit <NAME>",
+	Short: "Edit a cluster state",
+	Long:  `Use this command to edit a state.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) > 1 {
```
xmudrii: In case len of args is zero, (I think) we should try to read from env variable `KUBICORN_NAME`.

Such as in this case https://github.com/kris-nova/kubicorn/blob/master/cmd/apply.go#L47

cmd/edit.go
```diff
@@ -0,0 +1,140 @@
...
+
+var eo = &EditOptions{}
+
+var editCmd = &cobra.Command{
+	Use:   "edit <NAME>",
+	Short: "Edit a cluster state",
+	Long:  `Use this command to edit a state.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			ao.Name = strEnvDef("KUBICORN_NAME", "")
```
kris-nova: Should we default to an auto-generated name here instead of an empty string?

xmudrii: Huh I don't think so, but correct me if I'm wrong please. If I got this right, this will edit already defined cluster? So, changing random thing cloud be uh-oh. :D

cmd/edit.go
```diff
@@ -0,0 +1,140 @@
...
+			os.Exit(1)
+		}
+
+	},
+}
+
+func init() {
+	editCmd.Flags().StringVarP(&eo.StateStore, "state-store", "s", strEnvDef("KUBICORN_STATE_STORE", "fs"), "The state store type to use for the cluster")
+	editCmd.Flags().StringVarP(&eo.StateStorePath, "state-store-path", "S", strEnvDef("KUBICORN_STATE_STORE_PATH", "./_state"), "The state store path to use")
+	editCmd.Flags().StringVarP(&eo.Editor, "editor", "e", strEnvDef("KUBICORN_DEFAULT_EDITOR", "vi"), "The editor used to edit the state store")
```
kris-nova: Can we read from `$EDITOR` here? Most operating systems will already respect this value, so most users (myself included) will already have this set.. to `/usr/bin/emacs` in my case :stuck_out_tongue_winking_eye:

cmd/edit.go
```diff
@@ -0,0 +1,140 @@
...
+		os.Remove(fpath)
+		return err
+	}
+
+	cluster, err := stateStore.BytesToCluster(data)
+	if err != nil {
+		os.Remove(fpath)
+		return err
+	}
+
```
kris-nova: Can we hook into our init and validation here and ONLY `commit` the cluster if `initapi.InitCluster` passes?

cmd/edit.go
kris-nova: Ah good point - just saw the pattern somewhere else and thought about. No change needed here.

cmd/edit.go
```diff
@@ -0,0 +1,140 @@
...
+		os.Remove(fpath)
+		return err
+	}
+	err = cmd.Wait()
+	if err != nil {
+		logger.Debug("Error while editing. Error: %v", err)
+		os.Remove(fpath)
+		return err
+	} else {
+		logger.Info("Successfull edit")
```
xmudrii: Sorry for a very tiny nitpick, but it's `successful` :)

cmd/edit.go
MyDigitalLife: Didn't know that! Will change :)

PR #316, opened 2017-08-19
Updated docker file

MyDigitalLife:
Removed packages that are not needed anymore because there in the vendor.


docker/Dockerfile
```diff
@@ -2,6 +2,4 @@ FROM golang:1.8
 
 WORKDIR /go/src/
 
-RUN apt-get update && apt-get install -y libncurses5-dev
-RUN go get google.golang.org/api/compute/v1
-RUN go get cloud.google.com/go/compute/metadata
\ No newline at end of file
+RUN apt-get update && apt-get install -y libncurses5-dev
```
kris-nova: Do we still need this with #314 still?

xmudrii: Nope, it's not needed anymore.

MyDigitalLife: This was already removed in #314 so it should be gone when merged :)

prateekgogia:
I am not sure, if we even need this Dockerfile.

MyDigitalLife:
No it's not necessary at this point seeing the dependency was removed. It could be fully removed if we want to. But we could leave it there for a while seeing we might be adding dependencies in the future.

prateekgogia:
@MyDigitalLife Can we please remove the file for now. We can always add it later, if we need to add dependencies.

MyDigitalLife:
@prateekgogia Sure, I'll try and do this tonight.

PR #318, opened 2017-08-19
RFC: lint shell scripts with `shellcheck`

paulczar:
Initial pass of this adding the test and fixing a few shell scripts
to pass.  Not sure if its going overboard testing shell scripts, but
given we have a few I thought it would be a good idea.

### DIFF
--- a/scripts/headers.sh
+++ b/scripts/headers.sh
@@ -33,10 +33,11 @@ EOF
 FILES=$(find . -name "*.go" -not -path "./vendor/*")
 for FILE in $FILES; do
 	if [ "$FILE" == "./bootstrap/bootstrap.go" ]; then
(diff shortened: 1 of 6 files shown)


scripts/headers.sh
```diff
@@ -33,10 +33,11 @@ EOF
 FILES=$(find . -name "*.go" -not -path "./vendor/*")
 for FILE in $FILES; do
 	if [ "$FILE" == "./bootstrap/bootstrap.go" ]; then
```
prateekgogia: Indentation doesn't look right here.

paulczar:
thoughts on this folks?  do we want to try and get some lint/syntax testing in for all the bash scripts ?

PR #322, opened 2017-08-20
initapi: improve ssh key fingerprint generation

xmudrii:
This is a proposal for improving SSH Key Fingerprint generation in InitAPI. Therefore, it could be great if we could test this with multiple keys. I've tested this with SSH-RSA keys, and it works great, and much better as well.

In current implementation we have, we are getting Public Key Fingerprint from Private Key which doesn't seem needed. This PR, uses public key directly. They're not encrypted so we should have access to public key as it is. This way, we have much faster fingerprint generation and user doesn't need to enter password (InitAPI doesn't have SSH Agent available, so user have to enter it manually in every case).

Would like to hear what do you folks think.


kris-nova:
LGTM - but failing on the following... Maybe a rebase?

```12.42s$ make check-code
can't load package: package github.com/kris-nova/kubicorn/vendor/github.com/Azure/azure-sdk-for-go/datalake-store/filesystem: found packages filesystem (client.go) and datalakestore (version.go) in /home/travis/gopath/src/github.com/kris-nova/kubicorn/vendor/github.com/Azure/azure-sdk-for-go/datalake-store/filesystem
✖ gofmt needs to be run on the following files: 
/home/travis/gopath/src/github.com/kris-nova/kubicorn/cloud/amazon/public/resources/internetgateway.go
/home/travis/gopath/src/github.com/kris-nova/kubicorn/cloud/amazon/public/resources/internetgateway.go
/home/travis/gopath/src/github.com/kris-nova/kubicorn/cloud/amazon/public/resources/internetgateway.go
make: *** [check-code] Error 1```

xmudrii:
@kris-nova tests are passing now :green_apple:

Issue #321, opened 2017-08-20
Integration test for ASG


kris-nova:
Can we have the integration tests for AWS test that we can scale a cluster up/down as defined in #301

matyix:
@kris-nova I am happy to contribute this.

kris-nova:
Take it away!

matyix:
@kris-nova Submitted this [PR](https://github.com/kris-nova/kubicorn/pull/331) first to allow run tests at a considerable cheaper price.

PR #331, opened 2017-08-24
AWS Spot price support 

matyix:
This feature is adding the option to create the cluster using AWS spot instances. For `node` type instances (`cluster.ServerPoolTypeNode`) the launch configuration is created with a spot request and submitted. 

This feature is a potential pre-requisite for #321 in order to be able to run e2e nightly integration tests at a considerable cheaper price. 

There are several outcomes (handled): 

- Spot price is fulfilled and the instances are launched with the desired price
- Spot price can't be fulfilled: `Placed Spot instance request. Status Reason: Placed Spot instance request: sir-m6pi6f7p. Waiting for instance(s)`
- Spot price is not supported for the given instance type: `Placing Spot instance request. Status Reason: Value (t2.medium) for parameter instanceType is invalid. 't2.medium' is an unsupported instance type. Placing Spot instance request failed.`
- If a cluster can't be created and the end user doesn't want to wait any longer, the cluster can be deleted as usual and the bid is removed

**Limitations** 
```
The following instance types are not supported for Spot by AWS:

T2
HS1
```

Suggesting to use the `on-demand` instance price as the `spot-price` - this almost guarantees that the spot request is fulfilled but the instance is launchd with the current market spot bid price but never raises above the `on-demand` price. 

**Future improvements** 

- Add spot fleet support
- Add GCE `preemptible` VM support

### DIFF
--- a/apis/cluster/serverpool.go
+++ b/apis/cluster/serverpool.go
@@ -34,6 +34,7 @@ type ServerPool struct {
 	Name              string      `json:"name,omitempty"`
 	Image             string      `json:"image,omitempty"`
 	Size              string      `json:"size,omitempty"`
+	SpotPrice         string      `json:"spotPrice,omitempty"`
 	BootstrapScripts  []string    `json:"bootstrapScripts,omitempty"`
 	Subnets           []*Subnet   `json:"subnets,omitempty"`
 	Firewalls         []*Firewall `json:"firewalls,omitempty"`
--- a/cloud/amazon/public/resources/launchconfiguration.go
+++ b/cloud/amazon/public/resources/launchconfiguration.go
@@ -17,9 +17,11 @@ package resources
 import (
 	"encoding/base64"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/kris-nova/kubicorn/apis/cluster"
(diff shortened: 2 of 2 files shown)


apis/cluster/serverpool.go
```diff
@@ -34,6 +34,7 @@ type ServerPool struct {
 	Name              string      `json:"name,omitempty"`
 	Image             string      `json:"image,omitempty"`
 	Size              string      `json:"size,omitempty"`
+	SpotPrice         string      `json:"spotPrice,omitempty"`
```
kris-nova: So a change here is a change to the API.. Because we plan on refactoring the API anyway I think it's okay to start introducing changes like this now. But in the future it won't be nearly as easy.

My only reservation is that we might be introducing an AWS level API field into an API that is aimed at being provider agnostic. Maybe preface this field name with `Aws` such as `AwsSpotPrice` so users will know that this is only for AWS?

Regardless of if other clouds support a pricing change like this, I think calling it `AwsSpotPrice` for now works as this is the only implementation that uses it.

However this does beg the question of whether or not we want to have cloud objects in the long run..

kris-nova: On second thought we can just keep this as is, and have a validation check for passing it in with other clouds.

Do you mind adding a quick check to our validation?

matyix: Currently it's AWS only - but same can work with Google's `preemptible` support (will get there as well). 

I am currently away for a few days and not able to work on this - as far as I remember if you don't pass anything it's ignored. Can you/someone please add the validation meanwhile - before the master diverges too much :)

kris-nova: Let's merge and open a ticket! Enjoy your holiday!

kris-nova:
Opened #347

apis/cluster/serverpool.go
matyix: Thank you - @ganlaksh is volunteering to add the validation meanwhile for #347

matyix:
Thank you.

Issue #262, opened 2017-08-12
Update AWS Profiles to use t2.xlarge for masters


kris-nova:
From master after deploying a cluster on `AWS`

```
kris-nova:kubicorn kris$ k get po --all-namespaces
NAMESPACE     NAME                                        READY     STATUS              RESTARTS   AGE
default       nginx-4217019353-mx0vp                      0/1       ContainerCreating   0          1h
kube-system   calico-etcd-4j77w                           1/1       Running             0          1h
kube-system   calico-node-1mfrm                           2/2       Running             0          1h
kube-system   calico-node-mp2g7                           1/2       CrashLoopBackOff    23         1h
kube-system   calico-policy-controller-1727037546-56gnf   1/1       Running             14         1h
kube-system   etcd-ip-10-0-0-94                           1/1       Running             0          1h
kube-system   kube-apiserver-ip-10-0-0-94                 1/1       Running             0          1h
kube-system   kube-controller-manager-ip-10-0-0-94        1/1       Running             0          1h
kube-system   kube-dns-2425271678-kn460                   0/3       ContainerCreating   0          1h
kube-system   kube-proxy-2ph5n                            1/1       Running             0          1h
kube-system   kube-proxy-cd4qg                            1/1       Running             0          1h
kube-system   kube-scheduler-ip-10-0-0-94                 1/1       Running             0          1h
```

So nothing is scheduling..

We need to fix this and solve integration testing sooner than later so this stops happening - we can't have master broken like this 😞

bgeesaman:
It appears that calico-etcd is listening on the master's nodeIP (in my case 10.0.0.63), the calico config map is saying calico-etcd is on 10.96.x.x (the podIP), and the instance of calico-node on the worker cannot therefore connect to the endpoint http://10.96.x.x:6666.  I modified the calico-configmap to make the ETCD_ENDPOINT http://10.0.0.63:6666, and then killed the calico-node pod on the worker.  I then added tcp 6666 to the security group on master to allow the client to connect, and the calico-node pod stopped crashlooping and my test pods were able to then start on the worker.

I'm no kubeadm expert, but there appears to be a race condition of sorts that could be exacerbated by the fact that you're using t2.mediums which can get bogged down on cpu credits during the apt-get installs in your bootstrap scripts.

kris-nova:
@bgeesaman tested and confirmed - intermittent on `t2.medium`s... I recently bumped our DO base install because load average was well over 12 during bootstrap..

[...]

PR #332, opened 2017-08-25
Update aws profiles to use t2.xlarge for master 

ganlaksh:


matyix:
If this were to resolve #262 I am not sure that this is the right way - as concluded it that issue track it had nothing to do with the master instance type but it was a communication problem between the subnets. That one is fixed by #333. 

Nevertheless if we'd like to increase the size is fine (most of the people I assume they use their own setup anyway), though not sure what is the rationale behind.

ganlaksh:
@matyix Yes this was not the root cause.. Just felt it might not be a good idea to have t2.medium as the default with cpu and memory limits and something like t2.xlarge would be a good start.But as you said, I believe most people are going to eventually have their own custom profiles and fine tune it according to their needs.

Issue #334, opened 2017-08-26
We need to enable GitHub Pages


ellenkorbes:
Just do:

Settings tab > GitHub Pages section > Source: master branch /docs folder

You're gonna get a message saying ` Your site is published at http://URL/`, I need to know what that URL is so I can update the `_config.yml` within the website. That's so all the `{{ relative_url }}` links will work.

kris-nova:
Kubicorn.io

PR #335, opened 2017-08-26
Website

ellenkorbes:
Here's all the website stuff so that when Kris turns the key on #334 it's all ready (though as mentioned a slight fix to `_config.yml` might be required).

Any idea why the build failed at 9f1dada  there? I'm drawing a blank.


prateekgogia:
docs/LICENSE.txt 
There is a LICENSE.txt  file which contains a different license, than what we have in other .go files. Is this required or we can have the same license we have in .go files in this file ?

ellenkorbes:
No, my bad, feel free to delete or replace that.

Issue #324, opened 2017-08-21
Implement --set for overriding API objects during create/apply


kris-nova:
Borrowing the idea from [helm](https://github.com/kubernetes/helm) I would like for us to implement a `--set` option during the `create` and `apply` commands in kubicorn.

The `--set` flag would accept a query string, and if the query was valid would flip the values out at runtime.

For instance.. Given an API that looked like

```yaml
name: myCluster
kubernetesAPI:
    port: "443"
..etc
```

A user could run the following command

```bash
kubicorn apply myCluster --set kubernetesAPI.port="4433"
```

and expect that to apply the cluster with the new value for the Kubernetes API, and ultimately update the state store with the new value after it's completion.

matyix:
@kris-nova Like the idea - was going to ask something similar, like reusing a VPC. It's quite common that people already have their own network settings configured and would like to have the cluster deployed inside that. During create if a valid VPC name is submitted than instead of creating/or launching in the default one we can use that. Would this issue intended to address these kind of change/reuse as well?

xmudrii:
Will close and open new issue instead of this one.

PR #341, opened 2017-08-26
feature allow override of cluster settings

paulczar:
create `--override` cli option for create/apply commands to
allow user to override a cluster setting.

Addresses #324


paulczar:
Currently sort of works, but having to Capitalize the keys because its working with the struct directly, probably need to do some marshalling etc so we can use `name` instead of `Name` etc. 

```
$ bin/kubicorn create test3 --profile aws -o "Name=bacon,Location=eggs"
```

kris-nova:
Excellent work! And that library is perfect! Can we please update the project dep files with the new dependency? 

Also I'm wondering if we can follow the other kubernetes command line tool vernacular and use --set instead of override?

PR #349, opened 2017-08-28
Adding fabulous to all commands

kris-nova:


cmd/delete.go
```diff
@@ -47,6 +48,9 @@ After the delete is complete, the state store will be left in tact and could pot
 
 To delete the resource AND the API model in the state store, use --purge.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		if os.Getenv("KUBICORN_TRUECOLOR") != "" {
```
xmudrii: @kris-nova shouldn't we do something like we do in `root.go`, so we support both EnvVar as well as `-f` flag?
E.g. https://github.com/kris-nova/kubicorn/blob/master/cmd/root.go#L65
Also, I'm not sure should we change `os.Getenv` to `strEnvDef`, because of consistency.

PR #366, opened 2017-08-31
Attempting my first docs change

kris-nova:
@ellenkorbes can you give me a little bit of guidance here?

If I was to merge this.. what exactly would happen 😸


ellenkorbes:
What exactly would happen is:

- The link to the documentation at the navbar would break because they're hard-coded and the `Documentation` there is expecting a lower-cased readme file;
- The new README would not be rendered at all because it's lacking the required YAML header;
- All files with a doctype other than general/aws/azure/do/google would be rendered into html but not listed anymore because currently the documentation lists are built by iterating those five keywords.

If you have Jekyll installed you can always test things by going to the /docs folder and entering `bundle exec jekyll serve`. That'll get you a working copy of the website on `localhost:4000`.

I tried making it so that it would iterate through every possible doctype, but there were some snags (the reason I moved away from Jekyll in our other project) so I had to settle for this less-than-ideal possible solution. 

You might find it acceptable though, it loops through every single doc file, sorted by doctype, and lists the doctype next to the name. If we do go with this we might wanna clean up our doctype values, e.g. `do` → `Digital Ocean`.

screenshot_20170831_093744

I'm still a bit swamped today but I should have all day tomorrow to work on all the kubicorn.io issues that have been piling up.

Issue #289, opened 2017-08-16
Harden our bootstrap process and define where our support starts/stops.


prateekgogia:
We would like to redesign the way bootstrapping is working in `kubicorn` to use a server side staticly linked binary to bootstrap Kubernetes. By design `cloud-init` will hang for the binary to complete and will fail if the binary exits non-0.

#### The new way of bootstrapping

We will run a small shell script to download the binary based on arch. The binary will run with the injected configuration from the shell script.
Cloud init will hang until the binary completes.
The binary will run a gRPC server.
The local `kubicorn` process will run a gRPC client.
This is how data will be exchanged to/from the servers during bootstrap.
For private network topologies will need to tunnel through a bastion.

#### Support for shell scripts

We will **ONLY** support the shell scripts we bake into the `kubicorn` binary at compile time that live in `/bootstrap`. 

Once #41 is complete, a user can optionally append our stable logic with their own logic. Once a user runs their logic the guarantee is removed, and we will no longer support the cluster. This is for advanced users only.

PR #371, opened 2017-09-01
Experimental node classes API (WIP)

pipejakob:
I wasn't able to flesh this out as much as I would have liked before sending it out, but I'm about to be out of the office for the long weekend, and there's still value in early feedback.

A lot of this design was inspired by Tim (read: stolen and probably misunderstood). I've skipped past the definition of Cluster itself because I'm fairly sold on the idea of it being kubeadm's API verbatim, and that we should figure out how to drive kubeadm to the point where that's feasible. I'll come back to it in the future.

Robert had a great idea of the Cluster definition being something that could be kept reasonably in sync across clusters in completely different cloud providers or on-premise installations, and that all of the cloud-specific bits could be modeled purely in the nodes (which is the focus of this PR).

We had brainstormed on a whiteboard how you could achieve something akin to class polymorphism using Kubernetes API objects, and this particular approach uses three levels: NodePools, NodeClasses, and NodeBuilders, described inline.

Related, I've also captured some of my ideas for kubeadm bootstrapping here:
https://github.com/pipejakob/tbd/blob/master/README.md

I'm keen to find out if you think it could fit in with https://github.com/kris-nova/kubicorn/issues/289, or at least might be one piece of the puzzle. We have the same problem in [kubernetes-anywhere](https://github.com/kubernetes/kubernetes-anywhere/blob/master/phase1/gce/configure-vm-kubeadm.sh) of duplicating the same OS-specific bootstrapping bits to prepare for `kubeadm init`.

@kris-nova @roberthbailey

### DIFF
--- a/apis/experimental/node_classes/types.go
+++ b/apis/experimental/node_classes/types.go
@@ -1,0 +1,152 @@
+package node_classes
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/pkg/api"
+)
+
+// This should match kubeadm's API as closely as possible, so its definition is
+// uninteresting and not fleshed out.
+//
+// What is more important (for this PR) is the representation of nodes and how
+// they are linked to the concept of the Cluster.
+type Cluster struct {
+	metav1.TypeMeta
+	metav1.ObjectMeta
+
+	// ...
+}
+
+// This is the level that cloud-agnostic controllers would interact with. This
+// is where one could set the desired size/shape of a NodePool.
+type NodePool struct {
+	metav1.ObjectMeta
+
+	Spec NodePoolSpec
+	Status NodePoolStatus
+}
+
+type NodePoolSpec struct {
+	Name string // foo
+	Class string // name of NodeClass
+	Args string // JSON { ... } or ugly substruct with cloud-specific fields.
+	            // this is explained in NodeClass.
+	Size uint64
+	MinSize uint64
+	MaxSize uint64
+}
+
+type NodePoolStatus struct {
+	Phase NodePoolPhase
+	Message string
+	Reason string
+
+	// In addition to phase, we probably want a representation of in-flight
+	// cloud requests so we don't duplicate them in further runs of the
+	// controller loop. Once we decide to spin up a new node, we should
+	// make note of that because it may still take a few minutes for it to
+	// become healthy and register, and we don't want to keep triggering
+	// instance creation just because current < desired. This could be
+	// avoided if we always rely on an ASG/MIG, but that might not give us
+	// the control we want.
+	PendingNodeAdditions []string
+	PendingNodeDeletions []string
+}
+
+type NodePoolPhase string
+const (
+	NodePoolPending NodePoolPhase = "Pending"
+	NodePoolRunning NodePoolPhase = "Running"
+	NodePoolTerminated NodePoolPhase = "Terminated"
+	// ...
+)
+
+// These are objects that a cluster admin (or infrastructure admin, if a
+// differnet role) would create in order to set the policy of what kinds of
+// nodes are available and allowed for the cluster, analogous to
+// PersistentVolumes and PersistentVolumeClaims. They can have generic names
+// like "small," "medium," "gpu," or as arbitrarily complex as is useful
+// ("large-gpu-us-west-ssd"). In a cloud context, it allows infrastructure
+// admins to place restrictions on what resources are allowed to be created,
+// and in an on-premise context, allows accurate modeling of what hardware is
+// actually available for use.
+//
+// These serve as the glue between NodePools and NodeBuilders. In order to keep
+// the NodeClass registry from becoming a combinatorial explosion, we can
+// parameterize attributes that aren't core to the NodeClass' identity. For
+// instance, it might be very important that region/zone is a core part of a
+// NodeClass, and an admin wants their naming scheme to always start with the
+// zone. However, within that zone, the exact disk image you use is far more
+// flexible and inconsequential to the cluster admin, so it's modeled as an
+// overridable attribute. NodePools can then specify Args to fill in the blanks
+// or override defaults, and NodeBuilders will merge the Args into the Params
+// and use the result to populate a node's attributes.
+type NodeClass struct {
+	Name string
+	Params NodeClassParams
+	NodeBuilder api.ObjectReference
+}
+
+type NodeClassParams struct {
+	InstanceType string // n1-standard-1, r2.medium, etc.
+	Disks []Disk
+}
+
+// Open question: is there value in modeling key package versions for a node?
+// Can be represented in an OS-agnostic way a la cloud-init, but would let us
+// have much more control over version of Docker, kubelet, etc., and perform
+// rolling updates (potentially in-place).
+type Disk struct {
+	Bootable bool
+	Size uint64
+	Image string
+}
+
+// This would be attached to control plane components and NodeSpec so that you
+// can trigger upgrades.
+type KubernetesVersion struct {
+	SemanticVersion string
+	Path string // gs://..., s3://...
+}
+
+// NodeBuilders are the concrete cloud-provider definitions. They should have
+// enough configuration to be able to make the service calls necessary to
+// create/destroy/manipulate VMs, etc. The controller that ultimately makes a
+// call to AWS to ask for a new node would be watching the AWSNodeBuilder
+// objects and all NodePool objects that are linked to them in order to be able
+// to reconcile state.
+//
+// These would likely be created by the administrator (or cluster provisioning
+// tool) as a way of linking cloud service accounts and settings with the
+// cluster.
+//
+// There's an open question about how these should link to the Cluster object,
+// especially if we want to support representing N clusters as concurrent
+// objects (for usecases like Cluster Registry, Federation, etc.). One way is
+// to explicitly link to Cluster object via ObjectReference. Another way is to
+// always give the controller a label selector of which objects it should care
+// about reconciling, so it can be sliced and diced however the admin wants.
+
+type AWSNodeBuilder struct {
+	Name string
+	ServiceAccount string
+}
+
+type GCENodeBuilder struct {
+	Name string
+	Project string
+	Zone string
+}
+
+type GKENodeBuilder struct {
+	Name string
+	Project string
+	Zone string
+	ClusterID string
+}
+
+// AKA NoopNodeBuilder, although this could also know how to trigger a hardware
+// provisioning process, like we hint at with PersistentVolumeClaims.
+type OnPremiseNodeBuilder struct {
+	Name string
+}


apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+	Zone string
+}
+
+type GKENodeBuilder struct {
+	Name string
+	Project string
+	Zone string
+	ClusterID string
+}
+
```
kris-nova: Azure!

prateekgogia:
Thanks a lot @pipejakob for starting this effort. I have been thinking about this for a while now, a couple of things in my mind are - 

- In Cloud provider packages, currently we have `model` file, which models Kubicorn specific data object to Cloud Providers specific api object data. We can have a similar `de-modeler` file, which can convert the response from Cloud provider to Kubicorn object model.
- The logic for actual, delete rendering, apply can be a common separate package.

Its pretty vague currently in my mind, I will try to come up with some write up around the designs you have suggested. In broad sense, I currently see Kubicorn to provide a generic api across Cloud Provider infrastructure as well, something like this - 

image


Regarding Kubeadm bootstrapping, we definitely need something more robust, OS-agnostic and extensible. This will help with #289, as currently there is no feedback or any other mechanism to know when/ why bootstrapping failed. 

One of the name suggestion I have for naming the binary is PXE-kube (Pre-Kube Execution Environment), similar to PXE-boot.

apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/pkg/api"
+)
+
+// This should match kubeadm's API as closely as possible, so its definition is
+// uninteresting and not fleshed out.
+//
+// What is more important (for this PR) is the representation of nodes and how
+// they are linked to the concept of the Cluster.
+type Cluster struct {
```
roberthbailey: Do we want the top level cluster object to include both the node definitions and the control plane definition to make it easier for people to read in a yaml file? That would mean that we would add a ControlPlane struct and embed it along with an [] of node pools in the Cluster struct.

apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+// call to AWS to ask for a new node would be watching the AWSNodeBuilder
+// objects and all NodePool objects that are linked to them in order to be able
+// to reconcile state.
+//
+// These would likely be created by the administrator (or cluster provisioning
+// tool) as a way of linking cloud service accounts and settings with the
+// cluster.
+//
+// There's an open question about how these should link to the Cluster object,
+// especially if we want to support representing N clusters as concurrent
```
roberthbailey: I think that this API should punt on the N clusters case and focus on managing a single cluster. The registry can be the aggregation layer.

apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+type NodePoolPhase string
+const (
+	NodePoolPending NodePoolPhase = "Pending"
+	NodePoolRunning NodePoolPhase = "Running"
+	NodePoolTerminated NodePoolPhase = "Terminated"
+	// ...
+)
+
+// These are objects that a cluster admin (or infrastructure admin, if a
+// differnet role) would create in order to set the policy of what kinds of
```
zmerlynn: s/differnet/different/

apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+// to explicitly link to Cluster object via ObjectReference. Another way is to
+// always give the controller a label selector of which objects it should care
+// about reconciling, so it can be sliced and diced however the admin wants.
+
+type AWSNodeBuilder struct {
+	Name string
+	ServiceAccount string
+}
+
+type GCENodeBuilder struct {
```
zmerlynn: I don't know if we want to bother separate GCE and GKE at this level. They differ in control plane, obviously, but at the node level, they should look the same modulo where the master is and how to contact it. (I know this is just a sketch.)

apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+// they are linked to the concept of the Cluster.
+type Cluster struct {
+	metav1.TypeMeta
+	metav1.ObjectMeta
+
+	// ...
+}
+
+// This is the level that cloud-agnostic controllers would interact with. This
+// is where one could set the desired size/shape of a NodePool.
```
zmerlynn: I think this should provide more definition. In particular, a `NodePool` is a collection of homogeneous nodes that are all part of a Kubernetes cluster - it's something that should be deployable using a single IG/ASG.

apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+// or override defaults, and NodeBuilders will merge the Args into the Params
+// and use the result to populate a node's attributes.
+type NodeClass struct {
+	Name string
+	Params NodeClassParams
+	NodeBuilder api.ObjectReference
+}
+
+type NodeClassParams struct {
+	InstanceType string // n1-standard-1, r2.medium, etc.
```
zmerlynn: Do you want this to be cloud specific? The PV interface is a little more generic, last I looked. It seems more likely you'd want to e.g. specify a core count (and e.g. GPU count, or whether the hardware is dedicated (possible on some clouds), preemptible, etc.), though comparing these across cloud providers is perilous.

apis/experimental/node_classes/types.go
pipejakob: With your recent suggestion of scoping the Cluster API to be a single cluster, and letting the Cluster Registry be the de facto mechanism for tracking multiple clusters, I'm not sure it would offer mechanical value (since all node pools found on a cluster would correspond to that cluster). Do you think a multidoc yaml file is reasonable enough to get the full picture of a cluster without adding literal pointers in the structs?

What makes me hesitate here is that if someone wants to keep N different clusters reasonably in sync across different cloud providers or on-premise installations (using the same versions of Kubernetes, the same features, the same control plane flags, etc.), it would be nice if there were a limited set of API objects they try to keep completely in sync. This may not end up being feasible, but the fewer "ignore this particular field, which is very likely to differ from one cluster to the next" cases we have, the better I think. Linking the Cluster to the NodePools explicitly immediately gives us this case to ignore.

apis/experimental/node_classes/types.go
pipejakob: SGTM (until we have a good reason to want it).

apis/experimental/node_classes/types.go
pipejakob: That is a good requirement to state outright.

One of the other things we had discussed (but I haven't heard a strong desire for yet) was the possibility to have the semantics of a NodeTemplate instead of a NodePool. Any time you wanted a new node, it would copy all of the relevant attributes from the corresponding template, but you could update the template without necessarily updating all of the nodes that were created from it. After updating, new nodes created from the template would have the new attributes. That might make it easier to create one-off node types for experimentation without the overhead of always having to create a new NodePool of size 1, although we could also support orphaned nodes that weren't part of any group.

apis/experimental/node_classes/types.go
pipejakob: :-)

apis/experimental/node_classes/types.go
pipejakob: Ah, good point. This is definitely an area that could use more exploration and prototyping to see what actually makes sense.

apis/experimental/node_classes/types.go
pipejakob: Yeah, I'm on the fence about this. One part of me wants to say that GKE is the product that users care about, and the fact that GCE nodes are used behind the scenes is an implementation detail that users shouldn't care about, and we could swap them out for custom GKE VMs if we really wanted to. But, on the other hand, I'm sure plenty of users are reaching around behind the GKE API and messing with the VMs directly anyway, so might as well empower them (to the extent we can) instead of trying to obscure this.

Definitely a topic for discussion with the GKE API team :-).

apis/experimental/node_classes/types.go
```diff
@@ -0,0 +1,152 @@
...
+//
+// There's an open question about how these should link to the Cluster object,
+// especially if we want to support representing N clusters as concurrent
+// objects (for usecases like Cluster Registry, Federation, etc.). One way is
+// to explicitly link to Cluster object via ObjectReference. Another way is to
+// always give the controller a label selector of which objects it should care
+// about reconciling, so it can be sliced and diced however the admin wants.
+
+type AWSNodeBuilder struct {
+	Name string
```
prateekgogia: Is the Name field here different, than what's in NodeClass struct. 
I am assuming this is the name with which the VM will get created. In that case, does it have to be cloud specific ?

apis/experimental/node_classes/types.go
prateekgogia: >Do you think a multidoc yaml file is reasonable enough to get the full picture of a cluster without adding literal pointers in the structs?

Does that mean if we have a yaml file with N cluster, there can be <= N control plane struct definitions in the file. 

What I understand is a sample yaml file would look something like this : 

```
kind: Cluster
metadata:
    name: clusterA
spec:
    nodePool:
    . . . .

----- 
kind: ControlPlane
metadata:
    clusterName: ["clusterA"]
spec:
    . . . .

----- 
kind: Cluster
metadata:
    name: clusterB
spec:
    nodePool:
    . . . .

----- 
kind: Cluster
metadata:
    name: clusterC
spec:
    nodePool:
    . . . .

----- 
kind: ControlPlane
metadata:
    clusterName: ["clusterB", "clusterC"]
spec:
    . . . .

```

prateekgogia:
@pipejakob @roberthbailey Any updates on this ?

roberthbailey:
Update: Jacob has been sketching out a design locally based on lots of input (most recently from the cluster autoscaler team, which shook things up a bit) but it hasn't been pasted back here. It's being periodically sync'd to https://docs.google.com/document/d/1G2sqUQlOYsYX6w1qj0RReffaGXH4ig2rl3zsIzEFCGY/edit for comments.

prateekgogia:
@roberthbailey  Can you please provide access to this doc.

kris-nova:
@roberthbailey 

Did we ever get the repository set up to port this work over to the new repo?

roberthbailey:
@kris-nova 
We decided (based on a suggestion from brian grant) to use the kube-deploy repository. I'm creating a pull request to create a cluster-api subdirectory there where we can start dumping code. 

@prateekgogia
The doc is [here](https://docs.google.com/document/d/1G2sqUQlOYsYX6w1qj0RReffaGXH4ig2rl3zsIzEFCGY/edit#) (join the cluster lifecycle mailing list to get access if you haven't already). Can you join the cluster api breakout meeting tomorrow? I plan to review relevant parts of that doc and (time permitting) have @pipejakob share what we have so far on the api design.

pipejakob:
Closing this in favor of https://github.com/kubernetes/kube-deploy/pull/298

Issue #354, opened 2017-08-28
Handling cloud provider specific API changes


matyix:
Currently we have a lean and generic API - as we move towards supporting more and more use cases we hit the differences between cloud providers. 

An example is #331 where we introduced Spot price support (currently for AWS, but the GCP implementation will follow shortly) - and had an API change. A new feature I am about to introduce is IAM support (for AWS, and later for GCP) - it's the same as above and supported by/or differently by each/few providers. 

Is there a preferred way to move forward by keeping the API neat and clean but also support provider specific items? Any thoughts on this for the future? Thanks.

_I was trying to work around these and submit a few PRs in Kops for specific needs as such (like #https://github.com/kubernetes/kops/issues/2983) but ultimately found [Kubicorn](http://kubicorn.io) and never looked back ever since :)._

kris-nova:
Ugh this is such a great issue. I think about this all the time. The two main approaches I have been toying with..


## 1 Cloud specific structs 

In the API we could have nullable cloud specific configuration such as

```go
type Cluster struct {
    Name string
    // other fields
    AwsConfiguration *AwsConfiguration
    // other cloud configurations
}

type AwsConfiguration struct {
    SpotPrice bool
}
```

## 2 Make them optional 

```go
type Cluster struct {
    Name string
    // other fields
    AwsSpotPrice bool
    // other optional fields
}
```

## My opinion

Go with option 1. I think it allows users to start coding in a ton of features and offers flexibility for clouds to grow.

###### Caveat 

I know spot price is cross-cloud - just using it as an easy-to-grok example

MyDigitalLife:
I also like option 1.

matyix:
Sounds good. Will refactor the spot feature and consider option 1 (which I prefer as well) when adding IAM support.

matyix:
Closing as PR was merged.

PR #383, opened 2017-09-03
Externalize cloud specific settings from API 

matyix:
Re issue #354 the cloud provider settings are now externalized from the API (in this case the AWS `spot price` feature) and sets the way to move forward regarding cloud specific settings. 

@kris-nova and @MyDigitalLife - from a practical experience I believe that the cloud specific settings are better placed in the `ServerPool struct` instead of `Cluster struct` as we concluded in #354 - this makes room to apply different settings per hostgroups/serverpools - going forward with the spot price example instead of adding MasterSpotPrice/NodeSpotPrice or MasterIAM/NodeIAM it's better to just have SpotPrice and IAM in the serverpool and apply to the relevant pool. For example we don't run master with spot price but node only, whereas the two groups have two different IAM roles attached.

Please advise if you like it this way.

### DIFF
--- a/apis/cluster/serverpool.go
+++ b/apis/cluster/serverpool.go
@@ -27,15 +27,19 @@ const (
 type ServerPool struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
-	Identifier        string      `json:"identifier,omitempty"`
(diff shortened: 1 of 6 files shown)


matyix:
Hi @kris-nova - any feedback on this, can we go ahead with the assumption that it's OK this proposed change?

PR #387, opened 2017-09-06
Adjust the security group to allow master to node

bgeesaman:
Fixes a typo preventing the master from being able to hit the node (proxy, logs, exec)

### DIFF
--- a/profiles/centosamazon.go
+++ b/profiles/centosamazon.go
@@ -119,7 +119,7 @@ func NewCentosAmazonCluster(name string) *cluster.Cluster {
 							{
 								IngressFromPort: "0",
 								IngressToPort:   "65535",
-								IngressSource:   "10.0.100.0/24",
+								IngressSource:   "10.0.0.0/24",
 								IngressProtocol: "-1",
 							},
 						},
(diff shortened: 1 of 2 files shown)


matyix:
Was my copy/paste error. Since you update this can you add for CentOS profile as well in the same PR?

matyix:
Thanks @bgeesaman

bgeesaman:
@kris-nova FYI, this is partially breaking new AWS clusters for those that use the built-in profiles.  The cluster spins up and all pods run, but kubectl exec/logs/proxy/port-forward do not work until the security group is manually adjusted.

profiles/centosamazon.go
```diff
@@ -119,7 +119,7 @@ func NewCentosAmazonCluster(name string) *cluster.Cluster {
 							{
 								IngressFromPort: "0",
 								IngressToPort:   "65535",
-								IngressSource:   "10.0.100.0/24",
```
prateekgogia: I have a couple of questions regarding this - 

- I see the Kube nodes are created with this subnet - 10.0.100.0/24. How does this rule help traffic from nodes to master, or we don't need traffic from nodes to master ?

- What is listening on port 65535 in master ?

matyix: @prateekgogia the nodes are created within the specified CIDR - these mentioned above are firewall rules specifying the allowed source and range. Master is in 10.0.0.0/24 and allows ingress from Node which is 10.0.100.0/24 - based on the rule, and the other way around.
There is nothing listening in 65535, this is just the port range (from 0 to 65535 - due to dynamic ports you can't list them all) and -1 is the allowed protocols (TCP UDP ICMP).

PR #389, opened 2017-09-08
travis: add 1.x instead 1.8.3 as go version

dvrkps:


kris-nova:
Go does a fantastic job at (usually) not breaking backwards compatibility.. I would suggest we attempt to support `1.8` and every release after it until something goes wrong and we can't..

I assume that we won't run into any issues for quite some time however..

Also what does the `1.x` do exactly? Unsure what this PR is attempting to solve.

xmudrii:
‎@kris-nova `1.x` notation makes Travis use always latest Go version, currently 1.9.

dvrkps:
1.x match latest stable go version

PR #390, opened 2017-09-10
Make Amazon bootstrap read cluster API from file

boingram:
Update the amazon profiles to use jq to read the cluster API from a file as opposed to doing token replacement of the bootstrap scripts #183 

I extracted the shebang + enviro setup stuff into its own pre-script, generate the input redirection that writes the cluster as json into a file, and then concat that with the desired bootstrap scripts. 

I'll have Azure, Digital Ocean, and Google Cloud in future PRs.


kris-nova:
I think this is a good starting point, great job @boingram!

Let's get this in and take advantage of our pre-release cadence!

:smile:

PR #400, opened 2017-09-18
Fixed missing PublicKeyPath

MyDigitalLife:


cloud/digitalocean/droplet/resources/ssh.go
```diff
@@ -75,6 +75,7 @@ func (r *SSH) Actual(immutable *cluster.Cluster) (*cluster.Cluster, cloud.Resour
 				newResource.CloudID = strconv.Itoa(key.ID)
 				newResource.PublicKeyData = key.PublicKey
 				newResource.PublicKeyFingerprint = key.Fingerprint
+				newResource.PublicKeyPath = immutable.SSH.PublicKeyPath
```
xmudrii: I can't take a look into details, but shouldn't this use `key.PublicKeyPath` instead of `immutable`, same as the other things?

MyDigitalLife: `key.PublicKeyPath` does not exist. The keys are from your digital ocean account. It checks if you key in the immutable var is the same as one of the keys that you have in you DO account. When it finds it it sets the needed values. But without setting the public key path it can't create the private key path seeing that i uses the public key path to do that. 

So without this change it can't use the private key to create a connection if you already have your public key in you DO account.

xmudrii: Oh, it's the API answer, I wasn't paying attention. Sorry.

PR #402, opened 2017-09-20
[WIP] logger: use fabulous functions for creating new writer

xmudrii:
Work In Progress — do not merge
Depends on kris-nova/lolgopher#17

It will add support for using fabulous on Windows, with all that colors. :D 🌈


kris-nova:
Rainbows on Windows

Issue #396, opened 2017-09-15
Remove VPN from DigitalOcean bootstraping


xmudrii:
Currently, DO bootstrapping is very slow, mostly due to OpenVPN tasks.
As far as I know, OpenVPN is here to make Private Networking truly private. By implementation, DO Private Networking is possible to be reached from any other Droplet in the data center, including ones that we don't own.

I suggest totally removing VPN implementation and using [Cloud Firewalls](https://www.digitalocean.com/products/cloud-firewalls/) to protect Droplets. We would have following rules:
* Enable SSH and Kubeconfig for everyone
* Enable All TCP and All UDP traffic for Droplets tagged `cluster-name-master` and `cluster-name-node`.

That would make Private Networking actually private, as every other request would be blocked by the firewall.
That doesn't require any changes in bootstrapping process. We would have to change profile to use tags as Firewall Source.
With this, we wouldn't need `/tmp/.ip` file creating and downloading and bootstrapping would be way too faster.

If some additional layer of protection is still needed, we can add in WeaveNet, which would be very easy to implement.

xmudrii:
Take a look at #417

PR #407, opened 2017-09-23
do: allow firewalls to use tags

xmudrii:
Partially Addresses #396 

It will allow us to specify tags instead of IP addresses for DO firewalls.


matyix:
@xmudrii should not [this](https://github.com/kris-nova/kubicorn/blob/330655004fe3c9d232b45cce1eaf507d4229c5bb/profiles/ubuntudigitalocean.go#L119) be master or contain an ingress rule from `master` as well?

xmudrii:
Errr, it should be.Thank you very much, I'll correct the issue as soon as I can. :)

matyix:
@xmudrii Added #409 to quick fix it.

Issue #411, opened 2017-09-26
Hang during cluster delete (Resetting TimeOut counter.)


matyix:
PR #404 introduces some signal handling fixes and a new bug which made the cluster hang during delete. 

```
2017-09-25T16:43:45Z [▶]  Start deleting resource...
2017-09-25T16:43:45Z [▶]  subnet.Actual
2017-09-25T16:43:45Z [▶]  subnet.Render
2017-09-25T16:43:45Z [▶]  subnet.Delete
2017-09-25T16:43:45Z [✔]  Deleted subnet [subnet-5feb8038]
2017-09-25T16:43:45Z [▶]  subnet.Render
2017-09-25T16:43:45Z [▶]  Start deleting resource...
2017-09-25T16:43:45Z [▶]  securitygroup.Actual
2017-09-25T16:43:45Z [▶]  securitygroup.Render
2017-09-25T16:43:45Z [▶]  securitygroup.Delete
2017-09-25T16:43:45Z [✔]  Deleted Security Group [sg-8f5665f7]
2017-09-25T16:43:45Z [▶]  securitygroup.Render
2017-09-25T16:43:45Z [▶]  Start deleting resource...
2017-09-25T16:43:45Z [▶]  internetgateway.Actual
2017-09-25T16:43:45Z [▶]  internetgateway.Render
2017-09-25T16:43:45Z [▶]  internetgateway.Delete
2017-09-25T16:43:46Z [✔]  Deleted internetgateway [igw-6f7e730b]
2017-09-25T16:43:46Z [▶]  internetgateway.Render
2017-09-25T16:43:46Z [▶]  Start deleting resource...
2017-09-25T16:43:46Z [▶]  vpc.Actual
2017-09-25T16:43:46Z [▶]  vpc.Render
2017-09-25T16:43:46Z [▶]  vpc.Delete
2017-09-25T16:43:46Z [✔]  Deleted VPC [vpc-8b1b67ec]
2017-09-25T16:43:46Z [▶]  vpc.Render
2017-09-25T16:43:46Z [▶]  Start deleting resource...
2017-09-25T16:43:46Z [▶]  keypair.Actual
2017-09-25T16:43:46Z [▶]  keypair.Render
2017-09-25T16:43:46Z [▶]  keypair.Delete
2017-09-25T16:43:46Z [▶]  keypair.Render
2017-09-25T16:43:46Z [▶]  Resetting TimeOut counter.
```

xmudrii:
[...]

PR #412, opened 2017-09-26
Bugfix - Remove not needed channel drain

matyix:
The (unnecessary) channel drain caused Kubicorn to hang during cluster delete - see issue #411. This is a fix for the issue.


matyix:
@xmudrii @kris-nova thanks for the quick resolution/merge.

PR #422, opened 2017-10-01
Google network tags

fabriziopandini:
**Why we need this PR**
Currently, when creating an instance group/instance template on GCP, kubicorn applies the same network tags both to master and node groups, and this doesn't help if you plan to use network tags for defining firewall rules as much restrictive as possible. 

With this PR it will be possible e.g. to restrict the ingress traffic only to master nodes by defining a rule that applies only to the VMs with network tag "kubicorn-master"

**Notes for the reviewer** 
- While making the change, I restricted the existing label "http-servers" and "http-server" only to the master server group; however, I'm not sure if the "http-server" label makes sense because kubeadm init always use TLS...
- Ideally other labels could be useful, like e.g. the name of the cluster... if you are aware of some use case for this as well, let me know.
- the PR includes also a small improvement to the google walkthrough, informing users that it is necessary to create a firewall rule to get the things going.

### DIFF
--- a/cloud/google/compute/resources/instancegroup.go
+++ b/cloud/google/compute/resources/instancegroup.go
@@ -198,12 +198,20 @@ func (r *InstanceGroup) Apply(actual, expected cloud.Resource, immutable *cluste
 	}
 
 	tags := []string{}
-	if immutable.KubernetesAPI.Port == "443" {
-		tags = append(tags, "https-server")
+	if r.ServerPool.Type == cluster.ServerPoolTypeMaster {
+		if immutable.KubernetesAPI.Port == "443" {
+			tags = append(tags, "https-server")
+		}
+
+		if immutable.KubernetesAPI.Port == "80" {
+			tags = append(tags, "http-server")
+		}
+
+		tags = append(tags, "kubicorn-master")
 	}
 
-	if immutable.KubernetesAPI.Port == "80" {
-		tags = append(tags, "http-server")
+	if r.ServerPool.Type == cluster.ServerPoolTypeNode {
+		tags = append(tags, "kubicorn-node")
 	}
 
 	prefix := "https://www.googleapis.com/compute/v1/projects/" + immutable.CloudId
--- a/docs/_documentation/google-walkthrough.md
+++ b/docs/_documentation/google-walkthrough.md
@@ -60,6 +60,11 @@ $ ls -al ~/.ssh/id_rsa.pub
 -rw-------@ 1 mhausenblas  staff   754B 20 Mar 04:03 /Users/mhausenblas/.ssh/id_rsa.pub
 ```
 
+Finally, it is necessary to create a firewall rule for allowing ingress traffic to the API server, that is esposed by default on tcp:443 in above profile.
+You can use [this guide to create firewall rules](https://cloud.google.com/compute/docs/vpc/using-firewalls); in production systems,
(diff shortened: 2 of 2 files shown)


xmudrii:
Hi there, thank you very for your contribution!

Could we create the firewall using Google API, something similar to what we do for DigitalOcean? It would help users to easier start going on with the cluster and will minimize need for the web UI.

fabriziopandini:
@xmudrii  that changes the scope of the PR, but it makes sense even if I don't know in detail what are you doing for Digital Ocean. I will work on this ASAP 👍 

Some initial thoughts about implementing automatic management of firewall rules for GCP
- to implement the full lifecycle of firewall rules (insert/delete) we should create dedicated firewall rules per cluster e.g. 'kubicorn-{clustername}-allow-kubectl'
- to create firewall rules per cluster it is necessary to add an additional network tag that allows to identify VMs in a cluster,  e.g. 'kubicorn-cluster-{clustername}'

WDYT?

xmudrii:
@fabriziopandini This sounds good time. I don't mind if we change the scope of the PR.
Would you be able to work on it or do you want me to merge the PR and create the appropriate issue?

fabriziopandini:
Let me check the dig into this a little bit in the week-end and then I will tell you if it make sense to implement management of PR rules in this PR or not

kris-nova:
@xmudrii can you handle this one please?

xmudrii:
Okay, I'll be taking over this. IIRC, I've got an GCE account so I can test it. I'll not merge this until I start working on the automatic implementation, to prevent errors for users using latest versions.

fabriziopandini:
@xmudrii @kris-nova 
I had a look on implementing automatic creation of firewall rules for GCE, and the first topic that should be addressed are 
1) if the expectation is to make kubicorn capable to host any type of GCE firewall rules or only a subset of possible combinations (basically what we need now)
2) if/how the current api representation of firewall rules in kubicorn can host what considered in scope for 1) 

With regards to 1) my suggestion is to limit the scope only at the ingress rule that we need to make things work.

With regards to 2) it depends by how much do you want to embed in kubicorn api concepts that are potentially GCP specific (I don't know other cloud provider in details, so I can't support here). e.g.in GCP
-  firewall rules are not tied to any specific instance groups; instead firewall rules can define the criteria for identifying the set of machines to rule applies on (destinations); this is where the concept of network tags kicks-in
- rules are tied to a network (there can be many networks; kubicorn now use implicitily default network)
- there is support for ingress rules of type allow or type deny
- rules can have a priority
etc. see https://cloud.google.com/compute/docs/vpc/firewalls for more details

What do you think about?
PS. if this PR is merging, we can move this thread on a separated issue a then create a separate PR afterwards

xmudrii:
For now, I would just support basic rules. It would allow users to easily start, yet they can use API or gcloud, or whatever to add more complex ones.

Adding GCE specific stuff is not a problem. E.g. we already have DigitalOcean specific implementation, so having one for GCE is not a problem. 

Basically, I'm not sure will @kris-nova agree, but I'm trying to minimize user interaction as much as possible. Even just on basic level. In the beginning, we don't need anything complex, but at least having fully-configured, ready to use cluster would be epic. 

I'm on mobile so I can't look further. I'll merge this PR soon and we'll move conversation to another issue. I already #452‎
‎

PR #423, opened 2017-10-01
Remove instance templates on kubicorn delete over GCP

fabriziopandini:
**Why we need this PR**
Currently, when creating deleting a cluster on GCP, kubicorn does not delete instance templates.
This PR fix this.

**Notes for the reviewer** 
- Before deleting instance templates, it is necessary that the deletion of the referencing instance groups completes. To implement this wait loop I implemented the same wait loop already implemented in GetConfig.
- While making the change, I made also tsmall changes to allow the delete procedure to run without error even if resources are partially or totally already deleted (e.g. useful if you have to recover from an error).

### DIFF
--- a/cloud/google/compute/resources/instancegroup.go
+++ b/cloud/google/compute/resources/instancegroup.go
@@ -49,6 +49,10 @@ const (
 	MasterIPAttempts = 40
 	// MasterIPSleepSecondsPerAttempt specifies how much time should pass after a failed attempt to get the master IP.
 	MasterIPSleepSecondsPerAttempt = 3
+	// DeleteAttempts specifies the amount of retries are allowed when trying to delete instance templates.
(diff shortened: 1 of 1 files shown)


cloud/google/compute/resources/instancegroup.go
```diff
@@ -338,6 +349,26 @@ func (r *InstanceGroup) Delete(actual cloud.Resource, immutable *cluster.Cluster
 	return renderedCluster, actual, nil
 }
 
+const (
```
xmudrii: Maybe we should push it to the of the page, with other constants, so it's not confusing

xmudrii:
Looks good to me, thank you very much for your contribution!

fabriziopandini:
🎉 my first contribution at kubicorn

PR #428, opened 2017-10-08
Closes #399. Adding a prompt using the go-prompt library

### LINKED ISSUE #399: Implement go-prompt
To make the Kubicorn experience even better we can implement [go-promp](https://github.com/c-bata/go-prompt). Basically it will add auto completion and some other fancy stuff with not a lot of effect I think. 

[Implementation idea](https://github.com/kris-nova/kubicorn/issues/399#issuecomment-333397923):
>Let's implement `go-prompt` for non-Windows OS-es. It looks very good, and it would be awesome to have something like this for kubicorn.
Here's the my implementation idea:
>1) Add `prompt.go` file to [`cmd`](https://github.com/kris-nova/kubicorn/tree/master/cmd) package which will contain the code.
>2) Add `prompt_windows.go` to [`cmd`](https://github.com/kris-nova/kubicorn/tree/master/cmd) package which will show not available error for Windows.
>3) Instead of defining everything manually, use `cobra` to obtain all commands and flags and build structure like [this](https://github.com/c-bata/kube-prompt/blob/master/kube/completer.go#L31).
>Cobra has many functions for obtaining flags, commands and descriptions, such as [Commands](https://godoc.org/github.com/spf13/cobra#Command.Commands).
>Full cobra docs available @ [godoc](https://godoc.org/github.com/spf13/cobra)
>
>If you'd like to learn more about `kubicorn` project, you can read [The Project Walkthrough](http://kubicorn.io/documentation/project-walkthrough.html).
>If you have any question, comment on the issue, join us on the `#kubicorn` channel in the [Gophers Slack community](https://invite.slack.golangbridge.org/) or open a PR. We'll always do everything we can to get the PR though.

mohanarpit:
This library also auto-completes while in the kubicorn shell. Currently windows is not supported. 
Since this is my first contribution to a Golang project, I'd love some feedback on the code (especially if I haven't used an idiomatic way to do something)


cmd/prompt_windows.go
```diff
@@ -0,0 +1,42 @@
...
+allow you to run commands interactively from the shell.
+Currently this doesn't work on Windows systems.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		logger.Critical("Sorry. kubicorn prompt is not available for Windows machines")
+		os.Exit(1)
+	},
+}
+
+func init() {
+	if runtime.GOOS == "windows" {
```
xmudrii: Let's guard this under build tags. In that case, you would remove that `runtime` check and add the `// +build windows` at the top of the file. I believe this would be better solution.

Build tags docs - https://golang.org/pkg/go/build/
An example of non-Windows file - https://github.com/kris-nova/lolgopher/blob/master/output.go
An example of Windows file - https://github.com/kris-nova/lolgopher/blob/master/output.go

cmd/prompt.go
```diff
@@ -0,0 +1,167 @@
...
+	flagSuggestions = commandFlagSuggestions[command]
+
+	return prompt.FilterContains(flagSuggestions, strings.TrimLeft(args[l-1], "-"), true)
+}
+
+func init() {
+	initializePrompt()
+	// Currently go-prompt doesn't work on Windows machines. For windows machines
+	// return an error saying that it's unsupported.
+	if runtime.GOOS != "windows" {
```
xmudrii: If we go with the build tags, this check is not needed.

cmd/prompt.go
```diff
@@ -0,0 +1,167 @@
...
+func RunPrompt() error {
+	fmt.Println("Please input your kubicorn commands.")
+	defer fmt.Println("Bye!")
+
+	p := prompt.New(
+		executor,
+		completer,
+		prompt.OptionTitle("kubicorn: interactive kubicorn client"),
+		prompt.OptionPrefix(">> "),
+		prompt.OptionInputTextColor(prompt.Yellow))
```
xmudrii: Let's get get rid of setting input text color. This is not user-friendly for users using bright colors for terminal (e.g. users with white terminal will hardly see anything).

Without this, terminal will automatically handle the color.
I'm also open to another solution if you have.

Also, I prefer putting parenthesis in the new line, to improve readability, e.g.:
```go
p := prompt.New(
 		executor,
 		completer,
 		prompt.OptionTitle("kubicorn: interactive kubicorn client"),
 		prompt.OptionPrefix(">> "),
)
```

cmd/prompt.go
```diff
@@ -0,0 +1,167 @@
...
+	s = strings.TrimSpace(s)
+	if s == "" {
+		return
+	} else if s == "quit" || s == "exit" {
+		fmt.Println("Bye!")
+		os.Exit(0)
+		return
+	}
+
+	cmd := exec.Command("/bin/sh", "-c", "kubicorn "+s)
```
xmudrii: Not sure is this a needed change, but let's rename this to prevent sort of unwanted effects, as the package is called `cmd` as well.

cmd/prompt.go
```diff
@@ -0,0 +1,167 @@
...
+	ret := make([]string, 0, len(args))
+	for i := range args {
+		if !strings.HasPrefix(args[i], "-") {
+			ret = append(ret, args[i])
+		}
+	}
+	return ret
+}
+
+func initializePrompt() {
```
xmudrii: This approach has a problem, but let's leave it as it as, because it would require total refactor of `cmd` code base.

The problem is that you will not have completions for any command after `prompt` as it's still not initialized.

As I said, let's keep this as it is and open up the new issue once this PR is merged.

xmudrii:
@mohanarpit Welcome to the kubicorn project! I left you some comments. This works and looks great, great job!
Before being able to merge, we need to fix tests as well.

You have to vendor `github.com/c-bata/go-prompt`. We're using [`dep`](https://github.com/golang/dep) for dependency management and you can learn more about this process in our [contributing guidelines](https://github.com/kris-nova/kubicorn/blob/master/CONTRIBUTING.md#vendorning).

cmd/prompt.go
mohanarpit: This is required because once inside the prompt, this is the command that will get executed whenever the user pressed enter. Have renamed the variable to remove confusion here.

cmd/prompt.go
mohanarpit: Sorry, I don't completely understand the issue here. The initializePrompt is called in the init() function. This causes the prompt.go to initialize whenever kubicorn is invoked via `kubicorn prompt`. 
And when you say "refactor of cmd code base", does that mean refactoring only prompt.go or the way the entire package is organized.

mohanarpit:
I've vendored go-prompt and the required libraries in the vendor/ directory. I'm not sure what automated tests I can write here. Can you guide me to some example test that can be written for this change?

xmudrii:
Tests are not needed here, for cmd package. Looks good to me, but I'm away right now and will do the final review and merge once I get back.

mohanarpit:
@xmudrii Sure. Thanks a lot for the blazing fast response! 😄

Issue #417, opened 2017-09-28
Improve --set flag and implement it for the apply command


xmudrii:
Currently, we have `--set` flag only implemented for the `create` command. We would like to have it for `apply` command as well.

But there as an bug in current implementation, which should be fixed before. As reported in the issue #381, current implementation is case sensitive, so if you execute:
```bash
kubicorn create knova -p azure --set location=uswest
```
It would fail. But if you execute
```bash
kubicorn create knova -p azure --set Location=uswest
```
It would work. It should work as expected with lower case too.

The [`helm`](https://github.com/kubernetes/helm) project implements something similar, so maybe it could be useful to take a look at their implementation. Their implementation is located [here](https://github.com/kubernetes/helm/blob/master/cmd/helm/install.go#L137).

Our current implementation is in the [`cmd/create.go`](https://github.com/kris-nova/kubicorn/blob/master/cmd/create.go#L149) file.

If you'd like to learn more about `kubicorn` project, you can read [The Project Walkthrough](http://kubicorn.io/documentation/project-walkthrough.html).
If you have any question, comment on the issue, join us on the `#kubicorn` channel in the [Gophers Slack community](https://invite.slack.golangbridge.org/) or open a PR. We'll always do everything we can to get the PR though.

danicat:
@xmudrii the link to the project walkthrough in the issue description is broken (small typo on the `project` word) http://kubicorn.io/documentation/projetct-walkthrough.html -> http://kubicorn.io/documentation/project-walkthrough.html

m4ns0ur:
@xmudrii you are using `swalker` dependency, which is case sensitive.
I don't no what is the edge cases here, but an easy fix is changing [L153](https://github.com/kris-nova/kubicorn/blob/master/cmd/create.go#L153) to this:
```
err := swalker.Write(strings.Title(parts[0]), newCluster, parts[1])
```

xmudrii:
@m4ns0ur that could work, would you like to open a PR?

m4ns0ur:
@xmudrii sure! Is there any related test on this file, to be updated?

xmudrii:
@m4ns0ur nope, we don't have `cmd` related tests

PR #431, opened 2017-10-09
cmd: implement set flag for apply

xmudrii:
Addresses part of the #417 issue.
This PR implements `--set` logic for the `apply` command.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -27,6 +27,7 @@ import (
 	"github.com/kris-nova/kubicorn/state"
 	"github.com/kris-nova/kubicorn/state/fs"
 	"github.com/spf13/cobra"
+	"github.com/yuroyoro/swalker"
 )
 
 type ApplyOptions struct {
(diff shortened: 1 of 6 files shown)


cmd/apply.go
```diff
@@ -97,6 +98,17 @@ func RunApply(options *ApplyOptions) error {
 	}
 	logger.Info("Loaded cluster: %s", cluster.Name)
 
+	if options.Set != "" {
+		sets := strings.Split(options.Set, ",")
+		for _, set := range sets {
+			parts := strings.SplitN(set, "=", 2)
+			err := swalker.Write(strings.Title(parts[0]), cluster, parts[1])
```
prateekgogia: This will crash if I enter flag as =  set1=1,set2=2,

xmudrii: @prateekgogia will this do the job?

cmd/apply.go
```diff
@@ -97,6 +98,20 @@ func RunApply(options *ApplyOptions) error {
 	}
 	logger.Info("Loaded cluster: %s", cluster.Name)
 
+	if options.Set != "" {
+		sets := strings.Split(options.Set, ",")
+		for _, set := range sets {
+			parts := strings.SplitN(set, "=", 2)
```
kris-nova: We probably want to expand on this later and support more than just this syntax but for now it's okay!

Issue #372, opened 2017-09-01
Use `~/.aws/credentials` file for Amazon's credentials


Noirbot:
I'm running into an issue with using Kubicorn to set up a Kube cluster in AWS, due to the fact that the account I'm trying to set up the cluster in is one I don't have a specific key/secret for. Instead, I have a key/secret for one account, and then that account has permissions to assume a role in the account I want to create the cluster in.

Because of this, my AWS CLI credentials looks like this:

```
[default]
aws_access_key_id = XXXXXXXXXXXXXXXXXXX
aws_secret_access_key = XXXXXXXXXXXXXXXXXXXXXXX
[kube_account]
role_arn = arn:aws:iam::ACCOUNT_NUMBER:ROLE_ID
source_profile = default
```

It seems like it should be possible to add some sort of `--aws-profile` flag to relevant commands.

It would also be lovely if, instead of having to set environment variables for everything, Kubicorn could just read in from the AWS CLI credentials file at `~/.aws/credentials`

[Implementation idea](https://github.com/kris-nova/kubicorn/issues/372#issuecomment-333403771):
>To implement this, you'll have to create AWS parser package which will read `~/.aws/credentials` and convert it to some Go struct. The best place for that package is the [`cutil`](https://github.com/kris-nova/kubicorn/tree/master/cutil) package. To read file, you can use [`parser`](https://github.com/kris-nova/kubicorn/blob/master/cutil/parser/fileresource.go) package.
>
>As of the `--aws-profile` flag, you'll need to add it to the [`cmd/apply.go`](https://github.com/kris-nova/kubicorn/blob/master/cmd/apply.go) file.
>
>If you'd like to learn more about `kubicorn` project, you can read [The Project Walkthrough](http://kubicorn.io/documentation/project-walkthrough.html).
>If you have any question, comment on the issue, join us on the `#kubicorn` channel in the [Gophers Slack community](https://invite.slack.golangbridge.org/) or open a PR. We'll always do everything we can to get the PR though.

matyix:
Are you planning to work on this or it's more like a feature request? Asking because I have a sketch version for IAM roles already but needs tiding up before sending in a PR.

Noirbot:
This is more of a feature request from my end - I have little to no experience writing Go, but if I get some time, I'll try to look into it more.

metmajer:
+1

xmudrii:
[...]

PR #433, opened 2017-10-12
Fix for #372 - support full AWS SDK authentication (ENV, ~/.aws/crede…

matyix:
…ntials and instance profile roles)

There were 2 or 3 issues/asks recently (latest is #372) to support the full AWS authentication chain. By default AWS SDK support ENV, ~/.aws/credentials and instance profile based authentication (in this order) and does fallback until fails. This PR removes the short-circuit which forces Kubicorn to fail if the environment variables are missing. Also the SDK provides meaningful error messages in case none of the above are present.

### DIFF
--- a/cloud/amazon/awsSdkGo/sdk.go
+++ b/cloud/amazon/awsSdkGo/sdk.go
@@ -15,9 +15,6 @@
 package awsSdkGo
 
 import (
-	"fmt"
-	"os"
-
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


MyDigitalLife:
Can you also update the documentation referring to these env vars:
https://github.com/kris-nova/kubicorn/blob/master/docs/_documentation/envar.md
https://github.com/kris-nova/kubicorn/blob/master/docs/_documentation/aws-walkthrough.md

matyix:
Had to send it in a different PR  #434 - could not use git to squash commits together due to a poor internet connection.

PR #437, opened 2017-10-13
Build in Docker Container #119

### LINKED ISSUE #119: Build in Docker Container
Continuing the discussion from #116.

I like the idea of implementing builds in docker containers for a number of reasons:

* Standardized build environment
* Standardized build tools (see conversation in #69)
* Simplified build process for those who don't want to setup a full dev environment
* Portability for Travis/Jenkins (maybe beneficial?)

We should still use a Makefile and allow users to build with `make` locally on their machines but providing a Dockerfile which will establish a known working build environment automatically and run `make` for you and for the platform you choose via a command line argument, would be very useful for new users and users on platforms that do not have easy access to tools we may choose to use as part of the build process.

Ch3ck:
- [x] Fixes #119 
- [x] Add name to Authors file
- [x] Format fixes to bootstrap

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,42 @@
+FROM golang:latest
+
+RUN mkdir -p /go/src/github.com/kris-nova/kubicorn/
+
+WORKDIR /go/src/github.com/kris-nova/kubicorn/
+
+COPY apis		apis
+COPY cmd		cmd
+COPY Makefile	Makefile
+COPY state		state
+COPY profiles	profiles
+COPY test		test
+COPY bootstrap	bootstrap
+COPY docs		docs
+COPY vendor		vendor
+COPY cloud		cloud
+COPY cutil		cutil
+COPY examples	examples
+COPY scripts	scripts
+COPY Gopkg.lock Gopkg.lock
+COPY Gopkg.toml Gopkg.toml
+COPY main.go		.
+
+RUN CGO_ENABLED=0 GOOS=linux  make docker-build-linux-amd64
+
+FROM alpine:latest
+
+MAINTAINER Kris Nova <kris@nivenly.com>
+
+ENV PATH /go/bin:/usr/local/go/bin:$PATH
+ENV GOPATH /go
+
+RUN	apk add --no-cache \
+	ca-certificates
+
+WORKDIR /root/
+COPY --from=0 /go/src/github.com/kris-nova/kubicorn .
+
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM golang:latest
+
+RUN mkdir -p /go/src/github.com/kris-nova/kubicorn/
+
+WORKDIR /go/src/github.com/kris-nova/kubicorn/
+
+COPY apis		apis
```
xmudrii: I'm not really sure why we have bunch of that `COPY` commands? Wouldn't `go get` do the needed job?

Ch3ck: Yes @xmudrii that's true Let me try that, it should work.

Ch3ck:
@xmudrii Issue fixed, However I have added build support for the different os architectures. I think a possible update will be to make the user specify the particular build environment from darwin, freebsd, windows or linux making the default Linux. Just let me know your thoughts on this.

Makefile
```diff
@@ -16,6 +16,7 @@ all: default install
 
 compile: ## Create the kubicorn executable in the ./bin directory.
 	go build -o bin/kubicorn -ldflags "-X github.com/kris-nova/kubicorn/cmd.GitSha=${GIT_SHA} -X github.com/kris-nova/kubicorn/cmd.Version=${VERSION}" main.go
+	docker build -t kris-nova/kubicorn:v0.0.003 . #docker image build.
```
xmudrii: We should remove this from makefiles as it's assuming every user has Docker locally installed.

Makefile
```diff
@@ -26,6 +27,7 @@ bindata: ## Generate the bindata for the bootstrap scripts that are built into t
 	go-bindata -pkg bootstrap -o bootstrap/bootstrap.go bootstrap/ bootstrap/vpn
 
 build: authors clean build-linux-amd64 build-darwin-amd64 build-freebsd-amd64 build-windows-amd64
+	docker run --rm -v kris-nova/kubicorn:v0.0.003
```
xmudrii: Same here.

xmudrii:
@Ch3ck Can you please run the `make gofmt` to format all Go files? That should fix tests.

Ch3ck:
@xmudrii Done! Also, Do I need to refactor the docker/ directory since with a Dockerfile present that script is no longer necessary?

PR #439, opened 2017-10-14
Fix typo: conrtibuting -> contributing

jfrederickson:
Hey! I just started taking a look at kubicorn and spotted a typo while I was on the site. So here's a PR to fix it. :)

(At least I think so, anyway - considering the same text appears in multiple places it feels like there might be some autogeneration happening here.)


xmudrii:
Welcome to kubicorn project. Thank you very much for your contribution, LGTM! :)

PR #441, opened 2017-10-14
Adding the `--aws-profile` command line option to the creational commands.

### LINKED ISSUE #440: Add `--aws-profile` flag for selecting profile from the `~/.aws/credentials` file
In the PR #433, we've done several changes to allow usage of the `~/.aws/credentials` file. Parsing of the file is handled automatically, by the [`aws-sdk-go`](https://github.com/aws/aws-sdk-go). However, we have another problem now -- we can't specify the profile name, so it'll use the `default` profile every time.

[This document](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html) explains how you can configure the `aws-sdk-go` to use custom profile.

In order to fix this issue, you'll need to start with adding new flag in [`init function of apply.go file`](https://github.com/kris-nova/kubicorn/blob/master/cmd/apply.go#L65).

The rest of the implementation depends on the approach..
You can try adding `os.Setenv("AWS_PROFILE", flag_value)` to the `[Run function of Apply command](https://github.com/kris-nova/kubicorn/blob/master/cmd/apply.go#L46). If that works -- the PR is completed and you can create one.

If you'd like to learn more about `kubicorn` project, you can read [The Project Walkthrough](http://kubicorn.io/documentation/project-walkthrough.html).

fntlnz:
Referring to #440 
I preferred to use a different approach than setting the environment variable as proposed in the issue since the `Profile` option was already there in the Sdk.

I felt free to add a new parameter  to `cutil.GetReconciler` so that each adapter
can pass its own specific parameters needed at *runtime*.

In my opinion this is ideal since there's a whole kind of parameters that does not necessarily fits into the state, like in this case with the need to actually choose a profile from the `~/.aws/credentials` which is not related to the actual state but to the way the user want to run the whole thing at a specific moment.

Closes #440

Example:
```
  kubicorn apply --aws-profile default mycluster
```

Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -8,6 +8,7 @@ Eric Hole <eric.hole@gmail.com>
 fabriziopandini <fabrizio.pandini@gmail.com>
 Igor Vuk <parcijala@gmail.com>
 Joe Beda <joe.github@bedafamily.com>
+Jonathan Frederickson <jonathan@terracrypt.net>
 Joshua Carp <jm.carp@gmail.com>
 Kris Nova <kris@nivenly.com>
 Lachlan Evenson <lachlan.evenson@gmail.com>
--- a/cloud/amazon/awsSdkGo/sdk.go
+++ b/cloud/amazon/awsSdkGo/sdk.go
@@ -29,13 +29,14 @@ type Sdk struct {
 	ASG *autoscaling.AutoScaling
 }
 
-func NewSdk(region string) (*Sdk, error) {
+func NewSdk(region string, profile string) (*Sdk, error) {
 	sdk := &Sdk{}
 	session, err := session.NewSessionWithOptions(session.Options{
 		Config: aws.Config{Region: aws.String(region)},
 		// Support MFA when authing using assumed roles.
 		SharedConfigState:       session.SharedConfigEnable,
 		AssumeRoleTokenProvider: stscreds.StdinTokenProvider,
+		Profile:                 profile,
 	})
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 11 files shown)


fntlnz:
This can lead to a BC for whom uses `cutil.GetReconciler`

xmudrii:
Hi there, welcome to kubicorn project. This looks great, thank you very much for the contribution!

Auto-generated file (`bootstrap.go`) is not go formatted, so tests are failing. Can you please run `make gofmt` to format all Go files?

fntlnz:
@xmudrii formatted, committed and squashed

xmudrii:
@fntlnz Just to check a thing -- how this will work with existing environmental variables (`AWS_*`)? Will they continue to work or no?

fntlnz:
@xmudrii yeah there’s no change on that side, I just removed the tests that were checking for error if the env where not defined since it’s not the case anymore.

fntlnz:
One could even set the `AWS_PROFILE` and completely ignore this flag. This is worth mentioning in the docs too.


@xmudrii also to reply your last question, The Profile in the session strict is populated via the AwsProfile, it that is an empty string environment variables are used

xmudrii:
@fntlnz Thank you very much for the explanation! I can't merge this due to conflicts, so can you please resolve them and make sure to re-run `make gofmt` so we prevent tests problems?

fntlnz:
rebased

PR #443, opened 2017-10-14
docker: Refactored the docker build script.

Ch3ck:
Dockerfile added in #119 and PR #437 makes this script unnecessary


MyDigitalLife:
I don't think that the Dockerfile makes this script unnecessary. The reason I original wrote the script was so that you could run the script instead of needing to run a docker run command with the needed parameters, I'm lazy and having a script is just faster. This is still the case even with the Dockerfile existing.

Also looking at the Dockerfile I see [this](https://github.com/kris-nova/kubicorn/pull/437/files#diff-3254677a7917c6c01f55212f86c57fbfR9). With this the Dockerfile will always build using the master branch from github that will be fetched by the `go get` command. But the use of the script is to build the current branch with the changes from you have localy. 

I think we need to have a second look at what we want to do with this docker build process and how we want to implement it. Just my 2 cents.

xmudrii:
@Ch3ck @MyDigitalLife I think it should be better to leave the script as it is. That script doesn't affect anything and could only be helpful for someone in the future.

We still don't know how we want to play around Docker and Kubicorn. We have to decide how we want to use Docker and implement CI/CD system around it so we can easy test and ship new iterations.

Thank you very much for the contribution, but I think it would be better to close the PR. Once we get to finish the core features, we'll definitely come back with the implementation plan for Docker and Kubicorn.

Issue #453, opened 2017-10-18
Reorganize profiles


xmudrii:
Currently, [profiles](https://github.com/kris-nova/kubicorn/tree/master/profiles) are located all in the same directory, with format of `<os><provider>.go`.

That's not so easy to follow and as we add more stuff, it'll be hard to trace everything. 
We could split them in directories, such as:
```
/profiles/digitalocean/ubuntu_16_04
/profiles/azure/freebsd_11
```

If you're working on this one, make sure to update appropriate code and tests as well.
I know this is being used on at least two locations:
* [cmd/create.go](https://github.com/kris-nova/kubicorn/blob/master/cmd/create.go)
* [tests](https://github.com/kris-nova/kubicorn/tree/master/test)

If you'd like to learn more about `kubicorn` project, you can read [The Project Walkthrough](http://kubicorn.io/documentation/readme.html).
If you have any question, comment on the issue, join us on the `#kubicorn` channel in the [Gophers Slack community](https://invite.slack.golangbridge.org/) or open a PR. We'll always do everything we can to get the PR though.

gar:
I'll take this, if that's okay?

xmudrii:
Feel free to take it, sounds good to me

samanthakem:
@gar are u working on this still?

gar:
@samanthakem yes, and I hope to have a PR for it soon

gar:
@xmudrii I've just opened a PR for this issue. Let me know if you have any comments or suggestions for changes I should make. Thanks! :)

PR #457, opened 2017-10-19
Reorganise profiles into dirs for each provider

gar:
This PR is related to #453. It was actually a straightforward renaming and moving of files. Because all the functions in the profiles exist in the `profiles` package (both before and after the refactor), there were no code changes needed as far as I could see.

I tested this by building a local binary of `kubicorn` with the new profile layout and successfully created an Ubuntu cluster on AWS.

Unfortunately I wasn't able to get `make test` running successfully (even before making any code changes) so I wasn't able to verify that the tests also didn't need any code changes, but from reading the code this seems to be okay also.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -24,7 +24,10 @@ import (
 	"github.com/kris-nova/kubicorn/apis/cluster"
 	"github.com/kris-nova/kubicorn/cutil/logger"
 	"github.com/kris-nova/kubicorn/cutil/namer"
-	"github.com/kris-nova/kubicorn/profiles"
+	"github.com/kris-nova/kubicorn/profiles/amazon"
+	"github.com/kris-nova/kubicorn/profiles/azure"
+	"github.com/kris-nova/kubicorn/profiles/digitalocean"
+	"github.com/kris-nova/kubicorn/profiles/googlecompute"
 	"github.com/kris-nova/kubicorn/state"
 	"github.com/kris-nova/kubicorn/state/fs"
 	"github.com/kris-nova/kubicorn/state/jsonfs"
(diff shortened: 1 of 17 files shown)


xmudrii:
‎‎Hi, thank you very much for the PR. 

However, there are several problems. 

First one is Go convention. By style guide, it's recommended to call packages by directory name.
For example, this one: https://github.com/gar/kubicorn/blob/7db7a11de523b9d86cf51b49b3a6f39af7ba2563/profiles/digitalocean/ubuntu_16_04.go#L15
Package should be called `digitalocean` instead of `profiles`, because it's located in the digitalocean subdirectory. 

Second, I think I already mentioned this in the original issue, but you'll have to update other files as well. As far as I know there's `cmd/create.go` and `tests` directory that depends on `profiles` package. Make sure to change the package name (e.g. instead of `profiles.NewUbuntuDigitalOceanCluster`, it's `digitalocean.NewUbuntuDigitalOceanCluster` and import parts.
That should fix tests.

Optionally, you can rename functions. As package is already called `digitalocean` for example, we can drop `DigitalOcean` from structs/functions name (e.g. `NewUbuntuDigitalOceanCluster` to `NewUbuntuCluster`).

If you have any question, ping me. :)

gar:
@xmudrii  I've redone this with each profile in it's own package now and I've renamed the functions anywhere they were used. If there are any other issues with the PR just let me know and I'll fix them up!

The one thing I'm not sure about is when there was a "clash" in package names (e.g. `test/amazon/ubuntu_cluster_test.go` is in `package amazon` but needs to import the profile which is also in a `package amazon`). In this case I've just given the package an alias `profile`, which I hope is okay.

xmudrii:
@gar This looks very good. Thank you very much for the contribution! I agree on aliasing `profile` and it'll do the job for now.

Issue #310, opened 2017-08-18
kubicorn apply !!


kris-nova:
Can we please have kubicorn cache the most recent cluster created in a file on the local filesytem.. maybe in `~/.kubicorn` or something?

Then can we code in a feature into the `apply` cmd that will allow a user to use the Unix parlance `!!` after a `kubicorn apply` such as

```bash
kubicorn apply !!
```

xmudrii:
@kris-nova We can't use Unix parlance `!!` as it will be expanded to the previous executed command before command execution (on both Bash and ZSH). 
So if previous command was `kubicorn create test`, and you execute `kubicorn apply !!`, shell will execute `kubicorn apply kubicorn create test`, instead of `kubicorn apply !!`, therefore we can't check did someone provided `!!`.

That could be a good idea with some other symbol, if that makes sense.

PR #464, opened 2017-10-23
[WIP] Implement store state cache in ~/.kubicorn for #310

alex1x:
First part is to just save the state in `~/.kubicorn`. I think this should work for what was wanted in that issue (#310). Haven't implemented for the json as am looking for feedback first.

* Not sure about importing go-homedir... should I use the expandPath function instead? If so, how do I "bring" it to `fs/impl.go`? Looks like `apply.go` gets it from `create.go` but I'm not sure how.
* Should I implement any further logic in terms of when/how the state is cached?

Going forward I'll work with the ReadCachedState function to get the functionality that was envisioned as `kubicorn apply !!` -- since as was pointed out `!!` would be expanded to whatever was previously run in the shell, I'll try to use `kubicorn apply last-known-state` or `kubicorn apply --last-known-state` instead.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,13 +1,13 @@
 Ajit Kumar <kumaraji@google.com>
+Alex Tasioulis <tasioulis.alex@gmail.com>
 Arpit Mohan <me@arpitmohan.com>
 Bo Ingram <boingram6@gmail.com>
 Brad Pinter <brad.pinter@gmail.com>
 Carolyn Van Slyck <me@carolynvanslyck.com>
 Darron Froese <dfroese@salesforce.com>
-ellenkorbes <ellenkorbes@gmail.com>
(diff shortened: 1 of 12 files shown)


xmudrii:
This could work, looks good for now. 

About `go-homedir`.. I believe there's no need to vendor/use it. We should have functions doing so. At least, I'm sure we have it in `cmd/env.go`. If you need it, you can move it to `cutil` (make sure it doesn't exist there before moving).

Using the flag (`--last-known-state`) would be much better idea instead of `kubicorn apply last-known-state`. I don't believe somebody will use `last-known-state` as the name, but well.. we never know that. :D

kris-nova:
Closing as this is an older PR - please re-open if you'd like us to take another look! Thank you! <3

PR #465, opened 2017-10-24
Now AZs for both master/nodes subnets will be correctly set on cluster creation

jnardiello:
The current AWS implementation is broken when it comes to correctly define subnets in the configured AZs. By default the `create` command from the model defines `Location` in the subnets definitions but in `cloud/amazon/public/resources/subnet.go:96` we are setting `Zone:  r.ClusterSubnet.Zone`. 
`r.ClusterSubnet.Zone` is not defined which causes subnets to default to a random AZ.

One would think that the quick win is to simply change the model to set the `Zone` field instead of `Location`, unfortunately this isn't working as expected due to a race condition and shared structures state in `cloud/amazon/public/resources/subnet.go`. What happens here, starting at line `:48`, is that because `r.ClusterSubnet.Identifier` is empty on cluster creation, only the `else` condition is evaluated and `Zone` is left empty.

This 1 line change is the result of some serious debugging by me, [Gabriele](https://github.com/gabrielelana) and [Simone](https://github.com/siscia).

One important change introduced by this PR consists in breaking changes in the `yaml` definition where `ServerPools` will drop the field `Location` and have instead a `Zone` field. Still, we feel this was the intended original implementation.

### DIFF
--- a/cloud/amazon/public/resources/subnet.go
+++ b/cloud/amazon/public/resources/subnet.go
@@ -73,6 +73,7 @@ func (r *Subnet) Actual(immutable *cluster.Cluster) (*cluster.Cluster, cloud.Res
 		}
 	} else {
 		newResource.CIDR = r.ClusterSubnet.CIDR
+		newResource.Zone = r.ClusterSubnet.Zone
 	}
 
 	newCluster := r.immutableRender(newResource, immutable)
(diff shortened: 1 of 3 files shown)


xmudrii:
@jnardiello Thank you so much for the contribution and time spent investigating the issue! We really appreciate it.

I'll leave this PR for Kris, as she has way more experience with AWS than me.

/cc @kris-nova

PR #472, opened 2017-10-31
Add AWS instance profile support

matyix:
Add InstanceProfile feature for AWS. This enables Kubernetes features like: Automatic ELB allocation, PVC handling, etc.  

This is part of #418. 

- New model element: InstanceProfile
- New resource element: InstanceProfile
- Supports Add/Delete/Update InstanceProfile
- Add InstanceProfile to LaunchConfiguration if present
- Manage InstanceProfile - Role - Policy together
- Add Cloud specific configuration to Bootstrap scripts

PR #478, opened 2017-11-13
Adding IAM profile for node group and fixing kubelet service restart

matyix:
This is related to the previous IAM support #472 PR. Contains a fix and an example of an AWS profile for Ubuntu.


matyix:
Hi @kris-nova @xmudrii - for some odd reason the Travis job did not kicked off. Can you trigger it manually please, or shall I try to update the PR with a "fake" commit (then we do a squash merge)?

Thanks
Janos

xmudrii:
‎Better trigger it with false commit. I'll not have access to the computer for next several hours.

xmudrii:
Looks good to me. Thank you very much for all the great work! :)

PR #482, opened 2017-11-17
Adopt Cluster API and write controllers

kris-nova:
We're using the new cluster API! 😄 

 - Adopting new API and using `legacy` API as `providerConfig`
 - Adding anonymous cluster interface so we can hack into the rest of kubicorn
 - Writing controllers for node scaling 
 - Build process for controllers
 - Adding controller field to profiles 

https://github.com/kubernetes/kube-deploy/blob/master/cluster-api


cloud/amazon/public/resources/instanceprofile.go
```diff
@@ -185,10 +186,23 @@ func (r *InstanceProfile) Apply(actual, expected cloud.Resource, immutable *clus
 		logger.Debug("CreateInstanceProfile error: %v", err)
 		if err.(awserr.Error).Code() != iam.ErrCodeEntityAlreadyExistsException {
 			return nil, nil, err
+		}else {
+			profileinput := &iam.GetInstanceProfileInput{
+				InstanceProfileName: &expected.(*InstanceProfile).Name,
+			}
+			outInstanceProfile, err := Sdk.IAM.GetInstanceProfile(profileinput)
```
nrb: +1 to this - I was thinking of submitting a patch fixing this specifically.

kris-nova: please do I will rebase

kris-nova: I just rebased on master but if you want to PR this gopher it

stealthybox:
What is intended for state-stores?
Is `kubicorn apply` now going to sync the state-store into the Cluster-API?
In this case, the Cluster-API would be the authoritative storage.

What about in reverse? ( when Cluster-API state is newer than the state-store )
`kubicorn sync` ?
`kubicorn controller`?
I suppose the newer state store would win, and in the case of a conflict, we would trust the Cluster-API.

davidewatson:
@stealthybox: How can do you determine whether the state store is newer than the Cluster API objects?

@kris-nova: If we intend to allow users to modify Cluster API types directly, then do you think we should have a `k8s` state store for this case?

stealthybox:
@davidewatson probably using timestamps on the objects from the time of commit

kris-nova:
This PR will be abandon for favor of a different approach.

webwurst:
What approach will it be? Just curious.

kris-nova:
Closing in favor of #572

PR #486, opened 2017-11-21
Add Packet provider and first profile

deitch:
This PR adds a functional packet provider and an ubuntu based first profile. 

The `AUTHORS.md` was updated automatically when `make all` ran, not sure about all of the changes.

/cc @vielmetti

### DIFF
--- a/cloud/packet/packetSDK/sdk.go
+++ b/cloud/packet/packetSDK/sdk.go
@@ -12,22 +12,32 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package defaults
+package packetSDK
 
-import "github.com/kris-nova/kubicorn/apis/cluster"
+import (
+	"fmt"
+	"os"
 
-func NewClusterDefaults(base *cluster.Cluster) *cluster.Cluster {
-	new := &cluster.Cluster{
-		Name:          base.Name,
-		CloudId:       base.CloudId,
-		Cloud:         base.Cloud,
-		Location:      base.Location,
-		Network:       base.Network,
-		SSH:           base.SSH,
-		Values:        base.Values,
-		KubernetesAPI: base.KubernetesAPI,
-		ServerPools:   base.ServerPools,
-		Project:       base.Project,
+	"github.com/packethost/packngo"
+)
+
+// Sdk represents the client connection to the cloud provider SDK.
+type Sdk struct {
+	Client *packngo.Client
+}
+
+// NewSdk is used to create a Sdk client to connect to the cloud provider.
+func NewSdk() (*Sdk, error) {
+	sdk := &Sdk{}
+	apiToken := getToken()
+	if apiToken == "" {
+		return nil, fmt.Errorf("Empty $PACKET_APITOKEN")
 	}
-	return new
+	sdk.Client = packngo.NewClient("kubicorn", apiToken, nil)
+
+	return sdk, nil
+}
+
+func getToken() string {
+	return os.Getenv("PACKET_APITOKEN")
 }
--- a/cloud/packet/public/resources/device.go
+++ b/cloud/packet/public/resources/device.go
@@ -1,0 +1,361 @@
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package resources
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/kris-nova/kubicorn/apis/cluster"
+	"github.com/kris-nova/kubicorn/cloud"
+	"github.com/kris-nova/kubicorn/cutil/compare"
+	"github.com/kris-nova/kubicorn/cutil/defaults"
+	"github.com/kris-nova/kubicorn/cutil/logger"
+	"github.com/kris-nova/kubicorn/cutil/script"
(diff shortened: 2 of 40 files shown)


vielmetti:
Looks like Travis failed in "make check-headers"

deitch:
Oh yes, I see it. Hold on.

deitch:
Updated.

cloud/packet/packetSDK/sdk.go
```diff
@@ -0,0 +1,43 @@
+// Copyright © 2017 The Kubicorn Authors
```
xmudrii: We would appreciate tests here. [Example of Happy/Sad tests can be found here](https://github.com/kris-nova/kubicorn/blob/master/cloud/digitalocean/godoSdk/sdk_test.go).

cloud/packet/public/resources/device.go
```diff
@@ -0,0 +1,368 @@
...
+		newResource.Type = r.ServerPool.Size
+		newResource.OS = r.ServerPool.Image
+		newResource.Location = r.Location
+		newResource.BootstrapScripts = r.ServerPool.BootstrapScripts
+		newResource.SSHFingerprint = immutable.SSH.PublicKeyFingerprint
+		newResource.Name = r.ServerPool.Name
+		newResource.Tags = r.Tags
+	}
+
+	fmt.Println(immutable)
```
xmudrii: I don't believe that `fmt` is meant to be here.

cloud/packet/public/resources/device.go
```diff
@@ -0,0 +1,368 @@
...
+	logger.Debug("device.Actual newResource %v", newResource)
+
+	newCluster := r.immutableRender(newResource, immutable)
+	fmt.Println(newCluster)
```
xmudrii: Same here.

cloud/packet/public/resources/device.go
```diff
@@ -0,0 +1,368 @@
...
+	// take 2
+	logger.Debug("device.Apply project ID ready for processing [%s]", projectID)
+	if projectID == "" {
+		return nil, nil, fmt.Errorf("Cannot work without valid project ID")
+	}
+	// we can add the projectID to the expected
+	expectedResource.ProjectID = projectID
+	logger.Debug("device.Apply expected vs actual %v %v", *expectedResource, *actualResource)
+	j, _ := json.Marshal(*expectedResource)
+	fmt.Println(string(j))
```
xmudrii: Same here.

cloud/packet/public/resources/device.go
```diff
@@ -0,0 +1,368 @@
...
+	j, _ = json.Marshal(*actualResource)
+	fmt.Println(string(j))
```
xmudrii: Same here.

cloud/packet/public/resources/device.go
```diff
@@ -0,0 +1,368 @@
...
+	// Kubernetes API
+	// todo (@kris-nova) this is obviously not immutable
+	immutable.KubernetesAPI.Endpoint = ""
+
+	newResource := &Device{
+		OS:               deleteResource.OS,
+		Type:             deleteResource.Type,
+		Location:         deleteResource.Location,
+		BootstrapScripts: deleteResource.BootstrapScripts,
+		Count:            0,
```
xmudrii: Is `count: 0` indeed to be zero?

cloud/packet/public/resources/device.go
```diff
@@ -0,0 +1,368 @@
...
+		newResource.Type = device.Plan.Slug
+		newResource.OS = device.OS.Slug
+		newResource.Location = device.Facility.Code
+		newResource.BootstrapScripts = r.ServerPool.BootstrapScripts
+		newResource.SSHFingerprint = immutable.SSH.PublicKeyFingerprint
+		newResource.Tags = device.Tags
+		newResource.ProjectID = project.ID
+
+	} else {
+		newResource.Count = 0
```
xmudrii: Same here, is count expected to be zero?

xmudrii:
@deitch Thank you so much! We really appreciate it!
I've left you some comments that needs to be addressed before we can merge this. If you have any questions, ping me here on Slack.

Once we get this fix, I'll happily merge this PR.

cloud/packet/packetSDK/sdk.go
deitch: Sure thing. Will add.

cloud/packet/public/resources/device.go
deitch: Ugh! One issue I had when working through it, threw that in because of line and timing issues and never deleted. Nice catch...

deitch:
Removed all of those (awful) `fmt.Println()`. :-)

cloud/packet/public/resources/device.go
deitch: I think so, but am not sure. This is at the end of a successful `Delete()` operation. At this point, is the returned `Resource` expected to return:

* the _actual_ state of the cluster, In which case it should be `0`?
* the _configured_ state of the cluster, in which case it should be what was configured originally, but we happen to have `0` because they were deleted, in which case it should be `1` (or whatever was in the profile)?

Which one should be returned here?

cloud/packet/public/resources/device.go
deitch: Same question as with the delete. If `Actual()` runs and discovers that there are no devices that match the tag (e.g. we looked for `master` nodes and discovered there are none), then `len(valid)` will be `0`. In that case, should we not return how many we found, which is `0`?

xmudrii: This one makes sense, so leave it as it is.

cloud/packet/public/resources/device.go
xmudrii: Not really sure, but as far as I know, `Delete` returns actual state. Actual state is `Count: 0`, so it should stay zero. Thanks for checking with me.

deitch:
Added the requested test. Also updated the documentation to include a "packet-walkthrough.md" like the ones for Google, DO, AWS, etc.

I think this is it, but if I missed anything, let me know.

xmudrii:
Thank you so much for the hard work here!! 🎉 :shipit:

PR #487, opened 2017-11-23
Moved the sighandler from init to reconcile state handlers, fixing th…

matyix:
…e 10m exit if nothing happens after start. When using as a package it's extremely annoying that if there is no calls towards the reconciler Kubiciorn exits (the main app as well).


xmudrii:
Sounds good to me, thank you very much for the contribution. :shipit:

PR #494, opened 2017-11-26
[WIP] digitalocean: Replace OpenVPN with Weave

### LINKED ISSUE #415: Replace OpenVPN with WeaveNet in DigitalOcean implementation
This issue is in-place of the old issue #177.

One of the biggest reasons we implemented VPN for DigitalOcean is to additionally secure Private Networking. By implementation, every Droplet in the same Datacenter (even Droplets we don't own) can communicate with our Droplets via DO Private Networking. We want to prevent this and therefore additionally secure our cluster.

The problem with OpenVPN (which is currently implemented) is that it's too slow. Provisioning cluster can take up to 5-10 minutes, depending on the size of Droplets and many other factors.

I think this is unreasonable high time and we need to lower it down.

With PR #407 we limited traffic only to Droplets between themselves, so fixed the problem with DO Private Networking.

Let's fix the problem with OpenVPN by replacing it with WeaveNet.

To be able to get this working, you'll need to do the following:
1) Implement WeaveNet with Kubernetes by following [the official tutorial](https://www.weave.works/blog/weave-net-kubernetes-integration/).
As this requires Kubernetes to be up and running, we can't use [VPN Bootstraping](https://github.com/kris-nova/kubicorn/tree/master/bootstrap/vpn). You would have to update the master's Bootstraping file ([Ubuntu's](https://github.com/kris-nova/kubicorn/blob/master/bootstrap/digitalocean_k8s_ubuntu_16.04_master.sh) and [CentOS's](https://github.com/kris-nova/kubicorn/blob/master/bootstrap/digitalocean_k8s_centos_7_master.sh) files).
2) You'll need to disable VPN provisioning in [the reconciler](https://github.com/kris-nova/kubicorn/blob/master/cloud/digitalocean/droplet/resources/droplet.go#L122). The PR #410 has very awesome idea for handling this, by adding VPN field to the cluster's struct. You can take the implementation from PR if you want (if it doesn't get merged by then).

That would be everything needed to get it working.

If you'd like to learn more about `kubicorn` project, you can read [The Project Walkthrough](http://kubicorn.io/documentation/readme.html).

xmudrii:
Closes #415 

**WIP — Work In Progress**

### DIFF
--- a/cloud/digitalocean/droplet/resources/droplet.go
+++ b/cloud/digitalocean/droplet/resources/droplet.go
@@ -18,18 +18,14 @@ import (
 	"context"
 	"fmt"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/digitalocean/godo"
-	"github.com/kris-nova/klone/pkg/local"
 	"github.com/kris-nova/kubicorn/apis/cluster"
 	"github.com/kris-nova/kubicorn/cloud"
-	"github.com/kris-nova/kubicorn/cutil/agent"
 	"github.com/kris-nova/kubicorn/cutil/compare"
 	"github.com/kris-nova/kubicorn/cutil/defaults"
 	"github.com/kris-nova/kubicorn/cutil/logger"
-	"github.com/kris-nova/kubicorn/cutil/scp"
 	"github.com/kris-nova/kubicorn/cutil/script"
 )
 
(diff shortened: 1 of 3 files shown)


xmudrii:
Getting WeaveNet on DigitalOcean is quite hard out-of-box. I'll try to find new solution for this

Issue #496, opened 2017-11-28
kubicorn apply messes up existing kube config


deitch:
If I run `kubicorn apply`, it does the right thing by adding the cluster/user/context to my `~/.kube/.config`. 

But it does it by adding a brand-new entire set of yml to the end of the file. This means that I lose all of my existing unless I manually delete it. 

I don't know if it always does it, or if it has a hard time with certain structured kube config files.

xmudrii:
@deitch Thank you very much for reporting the issue. There're some reported problems with deleting kubeconfig already (#384). I'll look into it once I get some free time.

stealthybox:
Related to solving this issue: #110

davecaplinger:
Writing the config to a different local file path (which is what I think #500 allows?) seems like the wrong approach to fix this issue (in my opinion, of course).  For example, I have multiple cluster configs in `~/.kube/config` so I can easily switch with `kubectx` and as @deitch indicated, the new cluster info that kubicorn added got written to the right file; just not the right place in that file.  Further, the default context, cluster, and user names that kubicorn wrote conflicted with existing entries.

Since the config file has a only a flat list of users that are distinguished only by name, the names themselves are not cluster-specific, and if two contexts refer to the same user name, there's no way to determine which name is the correct one for that context.  So it would be bad to overwrite or duplicate the name used by some other cluster's context in the config file.

Since the name itself seems to be just a local label (the `server`, `key-data`, and `certificate-data` are the important bits), then the user, cluster, and context names that kubicorn uses could just replace the word "kubernetes" with "kubicorn-" plus the cluster name, such as `myfirstk8s: kubicorn-myfirstk8s-admin@kubicorn-myfirstk8s`.  That would also help ensure that if you're using kubicorn simultaneously for multiple clusters, changes to one profile don't overwrite the other as long as their cluster names are unique.

stealthybox:
Thanks @davecaplinger,
I think we can all agree on your suggested approach.

We recently vendored kubernetes/client-go in anticipation of this (as well as other efforts)
The current implementation literally copies kubeadm's kubeconfig, so we'll need to write one that is struct aware.

There are more linked issues here as well: https://github.com/kris-nova/kubicorn/issues/56

ashish-amarnath:
[...]

PR #500, opened 2017-12-05
new cluster annotation for kubeconfig local path

fabriziopandini:
Kubicorn by default appends to `~/.kube/.config` the kubeconfig file for newly created clusters.

This behaviours is more than fine when you work mainly with only one cluster, but it creates some problems when you work with many clusters at the same time (as reported in [#496](https://github.com/kris-nova/kubicorn/issues/496)).

This PR introduces a new alpha annotation at cluster level,  `cluster.alpha.kubicorn.io/kubeconfig-local-path`.

By setting this annotation in your cluster state, the kubeconfig file is saved instead in the specified path, allowing user to switch to one cluster to another e.g. by setting the `KUBENCOFIG` env var.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -24,6 +24,7 @@ import (
 	"github.com/kris-nova/kubicorn/cutil/agent"
 	"github.com/kris-nova/kubicorn/cutil/initapi"
 	"github.com/kris-nova/kubicorn/cutil/kubeconfig"
+	"github.com/kris-nova/kubicorn/cutil/local"
 	"github.com/kris-nova/kubicorn/cutil/logger"
 	"github.com/kris-nova/kubicorn/state"
 	"github.com/kris-nova/kubicorn/state/fs"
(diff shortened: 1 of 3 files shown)


cmd/apply.go
```diff
@@ -171,6 +171,10 @@ func RunApply(options *ApplyOptions) error {
 	}
 
 	logger.Always("The [%s] cluster has applied successfully!", newCluster.Name)
+	if path, ok := newCluster.Annotations[kubeconfig.ClusterAnnotationKubeconfigLocalPah]; ok {
+		logger.Always("To start using your cluster, you need to run")
+		logger.Always("  export KUBECONFIG='%s'", path)
```
nrb: Perhaps the file could be appended to `KUBECONFIG`?

fabriziopandini: @nrb the IMO value in the annotation should contain the file name
Should I rename from ClusterAnnotationKubeconfigLocalPath ClusterAnnotationKubeconfigLocalFile? 😉

stealthybox: I think what @nrb is asking for is to suggest the user do:
```bash
export KUBECONFIG="${KUBECONFIG}:%s"
```
`KUBECONFIG` is a list of hierarchical paths:
See: https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters#append-homekubeconfig-to-your-kubeconfig-environment-variable

cutil/kubeconfig/kubeconfig.go
```diff
@@ -28,6 +28,12 @@ import (
 	"golang.org/x/crypto/ssh"
 )
 
+const (
+	// ClusterAnnotationKubeconfigLocalPah is the annotation used on the cluster
+	// to specify an alternative local path for the cluster's kubeconfig file
+	ClusterAnnotationKubeconfigLocalPah = "cluster.alpha.kubicorn.io/kubeconfig-local-path"
```
stealthybox: quick typo:
`s/ClusterAnnotationKubeconfigLocalPah/ClusterAnnotationKubeconfigLocalPath/g`

cmd/apply.go
nrb: Yeah, what @stealthybox said. I might even go further and use `os.Setenv('KUBECONFIG', "${KUBECONFIG}:%s" and inform the user that the file was added.

fabriziopandini: Ok! I will update the PR asap

Thanks @nrb for the clarification
@nrb os.SetEnv _only_ can change the environment definition inside your process and in child process (not in the parent process)

nrb: Ahh, good catch. Yeah, in that case I'd support outputting the export command

fabriziopandini:
🎉

PR #502, opened 2017-12-08
AWS cloud provider support

nrb:
Adds cloud provider integration into the AWS profiles. If this approach works for others, then I can look at doing similar in at least GCP and Azure.

I have confirmed that dynamic PVCs are fulfilled with EBS volumes with this. I believe ELBs work too, but I haven't tested it yet.


kris-nova:
YAYAYYYYYYY

PR #505, opened 2017-12-10
Fixing bootstrap script

kris-nova:


xmudrii:
‎@kris-nova Looks like there's the same problem in the node script as well.

PR #507, opened 2017-12-18
Allow nodes to manage EC2 volumes

nrb:
In order for containers on nodes to use dynamic PVs, the nodes need to
have the permission to attach and detach volumes. The master has
"ec2:*", and thus doesn't need these permissions explicitly.


xmudrii:
@nrb Looks good, merged. :shipit: Should we do the same change for Amazon CentOS profile as well?

nrb:
@xmudrii Yeah, I overlooked that. Submitted #508 to cover CentOS.

PR #510, opened 2017-12-21
state,s3: implement s3 state store

### LINKED ISSUE #414: Implement S3/Spaces as State stores
It would be useful to have S3 and Spaces as possible State stores.
[An State Store is defined as the interface with the following functions...](https://github.com/kris-nova/kubicorn/blob/master/state/interface.go).
State Stores are located in the [`state`](https://github.com/kris-nova/kubicorn/tree/master/state) package.

Currently, we have two implementations which you can use as example:
* [YAML implementation](https://github.com/kris-nova/kubicorn/tree/master/state/fs),
* [JSON implementation](https://github.com/kris-nova/kubicorn/tree/master/state/jsonfs).

In case you need to parse file from the bucket, we have [`parser`](https://github.com/kris-nova/kubicorn/tree/master/cutil/parser) package for parsing stuff from the remote resources. As far as I'm aware, S3 is not implemented there, so if you depend on it, you'll have to implement it as well.

Also, you'll have to update the [`cmd`](https://github.com/kris-nova/kubicorn/tree/master/cmd) pacakge with new state store option (e.g. [here](https://github.com/kris-nova/kubicorn/blob/master/cmd/create.go#L170)).

If you'd like to learn more about `kubicorn` project, you can read [The Project Walkthrough](http://kubicorn.io/documentation/readme.html).

xmudrii:
Closes #414


xmudrii:
rebased

Issue #511, opened 2017-12-21
Integration/E2E Tests for Kubicorn using heptio/sonobuoy


xmudrii:
We should create E2E tests to make sure our code is working as intended always.
That will cost money as will run this stuff on the real providers.

@kris-nova's and my plan was to use heptio/sonobuoy as it's easy to run tests using it.
My plan of attack is:
* Create and Apply K8s Cluster using Kubicorn
* Apply Sonobuoy test file
* Wait for the results
* Download them using `client-go` and parse them.

I've already tried to implement this but I encountered several problems:
* Creating DO cluster using similar functions to the [example](https://github.com/kris-nova/kubicorn/blob/master/examples/digitalocean/basiccluster.go) one is not working as intended -- cluster has only master node.
* I don't know how to parse results once I have them.

The first problem is not a big one and I'll probably get it fixed myself.
The second problem is bigger one and I don't really know what's the format of Sonobuoy's results file. I would love some help here.

The [`e2e`](https://github.com/kris-nova/kubicorn/tree/e2e) branch has some work but I believe code is quite outdated at this point.

stealthybox:
I have some helpful notes on how to reproduce and parse Sonobuouy results in the readme for the kubeadm conformance submission:
https://github.com/cncf/k8s-conformance/tree/master/v1.9/kubeadm#reproduce-conformance-results

xmudrii:
@stealthybox Thank you so much! I'll take a look. I think this is what I've needed.
I'm planning to wrap up something soon, hopefully even tomorrow. I'll ping you if I need some help. :)

stealthybox:
sure thing -- lmk if you wanna pair on anything 👍

xmudrii:
Here's what I've thought of:

Create Go package (that would be under `kubicorn/test` or `kubicorn/e2e`) that contains logic for each provider. The procedure would be as usual:
* Create cluster, get reconciler and reconcile it.
* The bootstrap script would contain Sonobuoy YAML file.
* Check are all nodes up and ready.
* Get Pod's logs and wait for the last log (there's special one in case of Sonobouy).
* Untar and parse them.

Doesn't sound hard. I planned to use `client-go` for most of the stuff. But I have a problem, I don't know how to vendor it. [They say it doesn't work with `dep`](https://github.com/kubernetes/client-go/blob/master/INSTALL.md#dep-not-supported-yet), and I don't have idea how to get around this one.
[...]

PR #520, opened 2018-01-17
vendor: try to vendor client-go

xmudrii:
This PR tries to vendor the `client-go`. The `apimachinery` seems to be already vendored.


stealthybox:
This work is for reading sonobuouy logs. See #511

PR #522, opened 2018-01-18
vendor: client-go, apimachinery and api

xmudrii:
This PR vendors `client-go` version 6.0.0 and its dependencies. In #520 we vendored version 1.5, which is old and is not usable due to some unknown errors (probably apimachinery and api version mismatches).

/cc @kris-nova @stealthybox


stealthybox:
LGTM

stealthybox:
:+1:

PR #524, opened 2018-01-23
profiles: added debian stretch profile for aws

adnxn:
This change adds a sample debian profile as described in https://github.com/kris-nova/kubicorn/issues/421

### DIFF
--- a/bootstrap/amazon_k8s_debian_stretch_master.sh
+++ b/bootstrap/amazon_k8s_debian_stretch_master.sh
@@ -1,0 +1,77 @@
+# ------------------------------------------------------------------------------------------------------------------------
+# We are explicitly not using a templating language to inject the values as to encourage the user to limit their
+# use of templating logic in these files. By design all injected values should be able to be set at runtime,
+# and the shell script real work. If you need conditional logic, write it in bash or make another shell script.
+# ------------------------------------------------------------------------------------------------------------------------
+
+# Specify the Kubernetes version to use
+KUBERNETES_VERSION="1.9.0"
+
+apt-get update -y
+apt-get install -y \
+    socat \
+    cloud-utils \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    software-properties-common \
+    jq
+
+curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | sudo apt-key add -
+
+add-apt-repository \
+   "deb https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") \
+   $(lsb_release -cs) \
+   stable"
+
+apt-get update && apt-get install -y docker-ce=$(apt-cache madison docker-ce | grep 17.03 | head -1 | awk '{print $3}')
+
+curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+
+cat << EOF > "/etc/apt/sources.list.d/kubernetes.list"
+deb http://apt.kubernetes.io/ kubernetes-xenial main
+EOF
+
+systemctl enable docker
+systemctl start docker
+
+apt-get update -y
+apt-get install -y \
+    kubelet \
+    kubeadm=${KUBERNETES_VERSION}-00 \
+    kubectl
+
+PUBLICIP=$(ec2metadata --public-ipv4 | cut -d " " -f 2)
+PRIVATEIP=$(ec2metadata --local-ipv4 | cut -d " " -f 2)
+TOKEN=$(cat /etc/kubicorn/cluster.json | jq -r '.values.itemMap.INJECTEDTOKEN')
+PORT=$(cat /etc/kubicorn/cluster.json | jq -r '.values.itemMap.INJECTEDPORT | tonumber')
+
+mkdir -p /etc/kubicorn
+
+cat << EOF  > "/etc/kubicorn/kubeadm-config.yaml"
+apiVersion: kubeadm.k8s.io/v1alpha1
+kind: MasterConfiguration
+token: ${TOKEN}
+kubernetesVersion: ${KUBERNETES_VERSION}
+api:
+  advertiseAddress: ${PUBLICIP}
+  bindPort: ${PORT}
+apiServerCertSANs:
+- ${PUBLICIP}
+- ${PRIVATEIP}
+authorizationModes:
+- Node
+- RBAC
+EOF
+
+kubeadm reset
+kubeadm init --config /etc/kubicorn/kubeadm-config.yaml
+
+# Thanks Kelsey :)
+kubectl apply \
+  -f http://docs.projectcalico.org/v2.3/getting-started/kubernetes/installation/hosted/kubeadm/1.6/calico.yaml \
+  --kubeconfig /etc/kubernetes/admin.conf
+
+mkdir -p /home/admin/.kube
+cp /etc/kubernetes/admin.conf /home/admin/.kube/config
+chown -R admin:admin /home/admin/.kube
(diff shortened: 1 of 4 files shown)


bootstrap/amazon_k8s_debian_stretch_master.sh
```diff
@@ -0,0 +1,75 @@
+# ------------------------------------------------------------------------------------------------------------------------
+# We are explicitly not using a templating language to inject the values as to encourage the user to limit their
+# use of templating logic in these files. By design all injected values should be able to be set at runtime,
+# and the shell script real work. If you need conditional logic, write it in bash or make another shell script.
+# ------------------------------------------------------------------------------------------------------------------------
+
+# Specify the Kubernetes version to use
+KUBERNETES_VERSION="1.8.7"
```
xmudrii: I believe we should match Master and Node versions. Going with `1.9.0` on both should be good enough.

bootstrap/amazon_k8s_debian_stretch_node.sh
```diff
@@ -0,0 +1,45 @@
+# ------------------------------------------------------------------------------------------------------------------------
+# We are explicitly not using a templating language to inject the values as to encourage the user to limit their
+# use of templating logic in these files. By design all injected values should be able to be set at runtime,
+# and the shell script real work. If you need conditional logic, write it in bash or make another shell script.
+# ------------------------------------------------------------------------------------------------------------------------
+
+KUBERNETES_VERSION="1.9.0"
```
xmudrii: As explained for the master script.

bootstrap/amazon_k8s_debian_stretch_master.sh
```diff
@@ -0,0 +1,75 @@
...
+
+apt-get update -y
+apt-get install -y \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    software-properties-common \
+    jq
+
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
```
xmudrii: There's a command for adding Docker GPG on Debian, so I guess we should use that one:
```
curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | sudo apt-key add -
```

bootstrap/amazon_k8s_debian_stretch_node.sh
```diff
@@ -0,0 +1,45 @@
...
+
+apt-get update -y
+apt-get install -y \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    software-properties-common \
+    jq
+
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
```
xmudrii: Same here:
```
curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID")/gpg | sudo apt-key add -
```

bootstrap/amazon_k8s_debian_stretch_master.sh
```diff
@@ -0,0 +1,75 @@
...
+
+add-apt-repository \
+   "deb https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") \
+   $(lsb_release -cs) \
+   stable"
+
+apt-get update && apt-get install -y docker-ce=$(apt-cache madison docker-ce | grep 17.03 | head -1 | awk '{print $3}')
+apt-get update && apt-get install -y apt-transport-https
```
xmudrii: You have already installed it above, on the line 11, there's no need to run the install command again.

bootstrap/amazon_k8s_debian_stretch_node.sh
```diff
@@ -0,0 +1,45 @@
...
+
+add-apt-repository \
+   "deb https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") \
+   $(lsb_release -cs) \
+   stable"
+
+apt-get update && apt-get install -y docker-ce=$(apt-cache madison docker-ce | grep 17.03 | head -1 | awk '{print $3}')
+apt-get update && apt-get install -y apt-transport-https
```
xmudrii: Same here, we've installed `apt-transport-https` on the line 10.

bootstrap/amazon_k8s_debian_stretch_master.sh
```diff
@@ -0,0 +1,75 @@
...
+apt-get install -y \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    software-properties-common \
+    jq
+
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+
+add-apt-repository \
```
xmudrii: If you want to skip this part, you can go with `docker.io` package instead of adding repos and so. That seems to work for Ubuntu (should be checked for Debian).

bootstrap/amazon_k8s_debian_stretch_master.sh
```diff
@@ -0,0 +1,75 @@
...
+
+cat << EOF > "/etc/apt/sources.list.d/kubernetes.list"
+deb http://apt.kubernetes.io/ kubernetes-xenial main
+EOF
+
+systemctl enable docker
+systemctl start docker
+
+apt-get update -y
+apt-get install -y \
```
xmudrii: I'm not sure what recommended packages this will fetch, but other bootstrap scripts seem to install `socat`, `etables` and `cloud-utils` as well. If you can, check do we need them.

bootstrap/amazon_k8s_debian_stretch_node.sh
```diff
@@ -0,0 +1,45 @@
...
+
+curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+
+cat << EOF > "/etc/apt/sources.list.d/kubernetes.list"
+deb http://apt.kubernetes.io/ kubernetes-xenial main
+EOF
+
+apt-get update -y
+apt-get install -y kubelet kubeadm kubectl
```
xmudrii: If we need `socat` and other packages, add them here as well.

bootstrap/amazon_k8s_debian_stretch_master.sh
adnxn: hm this seems to be the recommended way from [kubernetes.io](https://kubernetes.io/docs/setup/independent/install-kubeadm/#installing-docker). ill stick to this, but ill change up the adding GPG command as per your recommendation.

bootstrap/amazon_k8s_debian_stretch_master.sh
adnxn: fixed.

bootstrap/amazon_k8s_debian_stretch_master.sh
adnxn: fixed.

bootstrap/amazon_k8s_debian_stretch_master.sh
adnxn: fixed.

bootstrap/amazon_k8s_debian_stretch_node.sh
adnxn: gotcha. leaving `1.9.0` for both.

bootstrap/amazon_k8s_debian_stretch_node.sh
adnxn: fixed.

bootstrap/amazon_k8s_debian_stretch_node.sh
adnxn: fixed.

bootstrap/amazon_k8s_debian_stretch_master.sh
adnxn: hm seems to work with `socat`, `cloud-utils`. i'm not completely sure on whether they are actually needed.

bootstrap/amazon_k8s_debian_stretch_node.sh
adnxn: same as above.

xmudrii:
LGTM! Thank you, great work!

If you're interested, you can do the same for DO and GCP. Bootstrap scripts should be the same (pay attention the user for `KUBECONFIG`), and you should be able to the same profiles as for Ubuntu (make sure to change OS).

Issue #276, opened 2017-08-14
Implement Digital Ocean Cloud Controller for DO clusters


kris-nova:
Can we please implement the following with digital ocean clusters:

https://github.com/digitalocean/digitalocean-cloud-controller-manager

xmudrii:
‎TIL this is official project. +1
I could be doing it once I finish with signals, if somebody doesn't take before me.

xmudrii:
I'm taking on this one

xmudrii:
This issue is fixed by https://github.com/kubicorn/bootstrap/pull/23

PR #527, opened 2018-01-26
digitalocean: rework digitalocean bootstrap scripts

### LINKED ISSUE #523: Add a flag or an environmental variable to choose between local and remote bootstrap script
Currently, when we apply a state, remote script (from GitHub) is used to reconcile the cluster. This is not always convenient, especially when modifying and testing scripts.

For example, we can add an environmental variable, such as `KUBICORN_DEVELOPMENT_MODE` and check for it when we parse state's bootstrap script in the `parser` package. If it's set, use the local script, if not, get it from the repository.

xmudrii:
This PR reworks the cluster provisioning process on DigitalOcean. This method provides a great base for further improvements, such as implementing DO CCM (#276), Dashboard or Heapster.

### Changelog

* [ADD] Added **Components** API structure for defining optional cluster components.
* [ADD] Added and documented the `KUBICORN_ENVIRONMENT` variable, used to instruct `kubicorn` to use local bootstrap scripts. Closes #523.
* [ADD] Create a Cloud Firewall while provisioning the CentOS clusters.
* [FIX] Wait before populating the state if `godo` returns empty string instead of an IP address.
* [FIX] Fix networking issues by adding the appropriate flag to the `kubelet` service.
* [CHANGE] Switch to WeaveNet for CNI.
* [CHANGE] VPN is disabled by default now. In order to enable it, you need to change the `.components.vpn` to `true` in the profile or state. We're already protected by the firewall, so VPN is not necessary. Also, DigitalOcean will lock down Private Networking to Droplets on the same account soon, ensuring clusters are even safer.
* [CHANGE] Use a configuration file for `kubeadm` instead of flags.

All integration tests are passing with this setup (tested using heptio/sonobuoy).

Credits to @alexellis for his [Kubernetes blog post](https://blog.alexellis.io/your-instant-kubernetes-cluster/), covering how to install WeaveNet on the DigitalOcean cluster. It helped me a lot to finally get Weave working on DO.

### DIFF
--- a/apis/cluster/cluster.go
+++ b/apis/cluster/cluster.go
@@ -41,6 +41,7 @@ type Cluster struct {
 	Values            *Values        `json:"values,omitempty"`
 	KubernetesAPI     *KubernetesAPI `json:"kubernetesAPI,omitempty"`
 	GroupIdentifier   string         `json:"groupIdentifier,omitempty"`
+	Components        *Components    `json:"components,omitempty"`
 }
 
 func NewCluster(name string) *Cluster {
--- a/apis/cluster/components.go
+++ b/apis/cluster/components.go
@@ -1,0 +1,19 @@
+// Copyright © 2017 The Kubicorn Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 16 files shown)


alexellis:
Glad I could help with this :+1: Kubicorn looks great!

PR #531, opened 2018-02-02
Fix a few documentation links to GitHub :)

samba:
Links into GitHub refs (for directories & code) were failing 404 due to missing some path bits.


xmudrii:
Thanks! LGTM :green_apple:

Issue #131, opened 2017-08-03
Configuration files for kubicorn


xmudrii:
Maybe it would not be a bad idea to allow using of configuration files for kubicorn. For now, we can change some of settings using environmental variables. Using config files would allow users to easily set default values to various flags and settings.

One of possible solutions would be to use [spf13/viper](https://github.com/spf13/viper) which supports both configuration files (in various formats) as well as environmental variables.

It should not be hard to add, if everyone agree, I could do it.

kris-nova:
So I think a config file will be absolutely necessary as we start building `kubicorn` into containers to run as controllers in a pod.. but I am wondering what we would *gain* from doing it today?

It seems like we might be introducing unnecessary complexity? What do you think @xmudrii?

xmudrii:
@kris-nova Whenever you think it's good time. Adding it will not change anything in workflow - env. variables have highest priority, than provided flags when executing command and lastly config stuff.
I think we can wait with it until we fix few other stuff.

kris-nova:
Marko,

I trust you to own this issue if you want to take care of it. We just need to make sure that we remember that `kubicorn` is intended to be a library

xmudrii:
I'm assigning this issue to myself. :)

nrb:
@kris-nova 

> building kubicorn into containers to run as controllers in a pod.

I know  there's #482 that mentions some of this, but could you expand on this a little bit? Is the intention to deploy kubicorn itself into the cluster?

stealthybox:
@nrb that's what I got out of the PR.

The main use-case for Cluster-API is to use k8s API objects to manage machines.
ie: `kubectl apply -f machine.yaml` would create a Machine object in kubernetes, and a controller would realize that object into an actual node that joins the cluster.

Figured `kubicorn controller` would run as a daemon, either inside or outside the cluster.

xmudrii:
@kris-nova @nrb @stealthybox I guess the idea for running `kubicorn` in a pod is something similar to [this](https://www.nivenly.com/i-ran-terraform-in-kubernetes/).

What I thought about this PR is to create ability for `kubicorn` to handle configuration files. The priority would be Flags > Env Vars > Configuration file. The plan was to use `spf13/viper` but for some reason that didn't worked out for me.

I'm currently focused on the another set of issues but I'm planning to revisit this one soon.

If you have any ideas for this one, LMK. :)

adnxn:
[...]

PR #536, opened 2018-02-06
vendor: vendor spf13/viper

xmudrii:
Vendor `spf13/viper`, so we start working on #131.


stealthybox:
v1 Tag looks good to pin.
Vendor contents look appropriate.

/lgtm 👍

PR #543, opened 2018-02-20
Removing all the azure things!

kris-nova:
Just tidying up the repo and removing the broken azure integration


nrb:
What's the motivation for removing Azure completely?

xmudrii:
@nrb Mostly because it's too hard to support it. I have not worked on this one, but it seems like the Azure API and Go wrapper are not working very well for now.

[Here's a comment by Kris](https://github.com/kris-nova/kubicorn/pull/327#issuecomment-346147238).

nrb:
Thanks @xmudrii.  Would it be more likely to stay if running on Azure had some measure of corporate support helping to make sure it worked?

(Please note I'm not committing here, but gathering information)

kris-nova:
@nrb I reached this conclusion while work at Microsoft. So I have little faith that 1) corporate backing would change anything or even if it did 2) probably won’t happen anytime soon.


The problem is with virtual machine scale sets. Those just flat out don’t work. Or at least they didn’t for the few months I tried to hack on them. 

Note we COULD create an implementation calling the virtual machine API directly as the machine API will solve the same problem virtual machine scale sets was supposed to solve.

2nd note that the VMSS problem was just the straw that broke the camels back. The ARM API simply fails something on the order of 70% of requests - so we are just biting off even more work that terraform has already solved.


Happy to talk more tomorrow if you’d like!

nrb:
@xmudrii Thanks for the link to the comment, I had missed that.

@kris-nova I appreciate the context! I think all of that is sound and don't really have an argument for keeping it in. I think my one suggestion is to add some of that context to the commit message here for posterity.

PR #546, opened 2018-02-22
 [library] enhancements; moved some (cmd) cli logic to (pkg) pkg/cli

alejandroEsc:
moved some cli logic to pkg/cli, updated tests: pkg/agent/agent_test.go (to be a bit more clean, and fix issues running tests more than once) and pkg/kubeconfig/kubeconfig_test.go because tests assumes that local machine is ssh-able and so terminates prematurely and fails (fix allows it to be skippable).

### DIFF
--- a/cloud/amazon/public/resources/securitygroup.go
+++ b/cloud/amazon/public/resources/securitygroup.go
@@ -301,7 +301,7 @@ func (r *SecurityGroup) immutableRender(newResource cloud.Resource, inaccurateCl
 func strToInt(s string) (int, error) {
 	i, err := strconv.Atoi(s)
 	if err != nil {
-		return 0, fmt.Errorf("failed to convert string to int err: ", err)
+		return 0, fmt.Errorf("failed to convert string to int err: %#v", err)
 	}
 	return i, nil
 }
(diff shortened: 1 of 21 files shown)


xmudrii:
@alejandroEsc Seems good to me, thank you very much the contribution! Can you please fix the tests, some of them are failing.

alejandroEsc:
@xmudrii could you help me determine what failed? cant seem to find the particular case based on the logs, almost seems like the actual container failed? If so can we rerun the test?

xmudrii:
@alejandroEsc seems like that's not related to the PR, but for some reason fails.
Have you tried rebasing?

Anyways the error is in the [`cloud/amazon/public/resources/securitygroup.go`, on the line 304](https://github.com/kris-nova/kubicorn/blob/master/cloud/amazon/public/resources/securitygroup.go#L304).

The `errorf` call is missing formatting, it should be something along:
```
return 0, fmt.Errorf("failed to convert string to int err: %#v", err)
```

If rebasing doesn't help, fix that line and tests should start passing again.

alejandroEsc:
@xmudrii rebasing was clean, so opted to fix that line, as you mentioned.

xmudrii:
@alejandroEsc 🍏 

That looks good to me, I think it's okay to merge the PR.
Thanks a lot for the contribution and going through the process with me! We really appreciate it. :)

PR #547, opened 2018-02-23
Use an existing instance profile if found

nrb:
Since we can detect whether or not an instance profile exists via error
types, attempt to re-use an existing instance profile before returning
an error.

The name and id have to be extracted seperately since the returned
values for creating an instance profile and getting one are two
different types.

Signed-off-by: Nolan Brubaker <nolan@heptio.com>


nrb:
Rebased, waiting on Travis

xmudrii:
@nrb looks good to me. you can merge it once ci is :green_apple:

Issue #105, opened 2017-07-23
Implement Openstack


shwsun:
Hey, 
Unfortunately I don't have access to EC2 or Azure, it is possible to create some sort of profile for Openstack environment? If so I would be happy to help (I work on openstack a lot)

kris-nova:
Yes this is a great idea, would love to code an implementation in! There is documentation on how write a Reconciler in the Cloud directory https://github.com/kris-nova/kubicorn/tree/master/cloud and there are examples of other implementations in the code base. 

Do you want to start coding and we can work together? Also happy to jump on a video call and go through this in greater detail if needed

shwsun:
Hi,
I will be a little busy the coming week due to some research workload so that spending non-preempted time might be a little luxury for me :(  I will start with the reconciler and probably discuss with you if I have any questions. I am very interested into the bare metal case too (some of my past research can actually use that right away)

kris-nova:
I am wrapping up writing a book this week and next, but once that is done I should be back to my old tricks again.. meaning I can help out with this effort as well!

joedborg:
I'm happy to help with this too.  Closest thing, that I'm aware of, is Juju.

xlucas:
Has any progress be made on this subject ? Are you still looking for help ?

xlucas:
I've started working on an Openstack implementation.

xlucas:
I guess this issue can be closed now :)

xmudrii:
@xlucas thanks for the update! Closing

PR #549, opened 2018-02-23
OVH Cloud (Openstack)

xlucas:
Hello there!

I've been working during the past days on providing *an* Openstack cloud implementation (see #105). Openstack is used by many operators to provide public and private clouds.

However, there's actually no such thing as *the* Openstack implementation. Operators usually differ from one another by the collection of services they provide or not, by the way some of this components behave, and also by having different regions, different flavors, different images, different network topologies, and the list goes on.

For this reason, I thought it would be particularly well fitted to have:
- A generic set of Openstack resources (keypair, network, subnet, security groups and so on)
- Operator-bound profile and model (as well as resources if any), where specificities can be expressed.

Working myself at OVH, a large European cloud provider  which delivers a Public Cloud infrastructure based on Openstack, here's the first Openstack profile for Kubicorn !


xmudrii:
@xlucas Thank you much for such a great PR! Very well done! I've left you a comment—you need to update headers for new files.

Once tests are sorted, I think it's okay to merge this.

xlucas:
@xmudrii Actually the headers are already there. Looks like the check fails because the year is 2018 and not 2017.

xmudrii:
@xlucas yep. maybe it doesn't make sense at all, but script assumes it's 2017. Let's stay with 2017 for now. I can check with @kris-nova can we update headers.

xmudrii:
@xlucas Can you please rebase against the master branch? Maybe you want to look into the #546 to see more about the latest changes.

xlucas:
@xmudrii Done

xmudrii:
@xlucas A tiny nitpick. Personally, I don't prefer underlines in the Go file names, beside for test files, etc. For example, I prefer `securitygroup.go` over `security_group.go`.

I think it's the only occurrence. It's not a big thing and we can go without it, but maybe it's good idea to rename the file. In case for the profile file, I think it can stay, as other profiles are using it as well (sometimes, we're not consistent at all :P).

As said on Slack, if nobody takes on in the meanwhile, I'll do the final review tomorrow.

xlucas:
Well, I did some renaming.

cloud/openstack/openstackSdk/sdk.go
```diff
@@ -0,0 +1,83 @@
...
+// Sdk represents an Openstack SDK.
+type Sdk struct {
+	Compute       *gophercloud.ServiceClient
+	Network       *gophercloud.ServiceClient
+	ObjectStorage *gophercloud.ServiceClient
+}
+
+// NewSdk constructs a new Openstack SDK for the specified region.
+//
+// The following environment variable list is looked up in the user
```
xmudrii: Can we please document this. Those variables should be added to the [`envar.md`](https://github.com/kris-nova/kubicorn/blob/master/docs/_documentation/envar.md) file.

xmudrii:
@xlucas Sorry for the delay! I was very busy for the past few days.
Can we just document this? Beside that, this is very awesome. Thank you so much for such a great job!

xmudrii:
:shipit: once :green_apple: :)

xmudrii:
Thank you so much for all the great work! We really appreciate it! :shipit:

PR #552, opened 2018-02-28
Documentation improvement

xlucas:
I've spotted some jekyll layouts to be updated following #549.

Also, I propose to drop `_site` from git since it's a bit annoying when you are serving documentation on localhost to preview your changes.

However, it will require all contributors to run the following after pulling master to *completely* ignore these files in their fork:

```bash
git rm -r --cached docs/_site
git rm -r --cached docs/.sass-cache
```


xmudrii:
LGTM. Thanks for the contribution! :) :shipit:

PR #557, opened 2018-03-02
Update references from kris-nova/kubicorn to kubicorn/kubicorn

philoserf:
- the **kubicorn** has a new home. 🦄


philoserf:
Looking at failures. I do not think they are related to this PR. Still, taking a closer look.

- vet failures
- test failures
- AUTHORS drift

PR #566, opened 2018-03-08
ci: fix ci tests

xmudrii:


detiber:
It looks like the agent tests are throwing false positives still. How we spin up the test_ssh container still needs to be determined.

PR #568, opened 2018-03-09
Attempt to get ssh test working in CircleCI

detiber:


detiber:
@xmudrii @kris-nova It seems a bit odd to rely on an external service for ci/unit tests. It seems like more of an integration test, and I'm wondering if it would make sense to spin up an ssh server as part of the test setup.

I'm also wondering if this should also be the case for the minio setup as well. If for no other reason, then we could pick, test, and use random ports during test spin up and not have to worry about host port collisions and ensuring port configuration between the service container for the test and the test configuration.

 like we do for the ssh test. Would it make more sense if we spun up an ssh server as part of the test instead of relying on an external container?

At the very least thi

PR #542, opened 2018-02-17
Working on SSH

kris-nova:
Fixing #539

PR #571, opened 2018-03-10
add the key to the agent only once so passphrase is only once needed

ulrichSchreiner:
When using a ssh-key with a passphrase the loop in `RetryGetConfig` always creates new agents and always add the key-file to the agent. This forces one to always enter the passphrase in every loop iteration. This PR fixes this by creating the ssh configuration once and using this configuration for the whole loop. So a passphrase must be entered only once.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,6 +18,7 @@ fabriziopandini <fabrizio.pandini@gmail.com>
 Gar Morley <garmorley@gmail.com>
 Igor Vuk <parcijala@gmail.com>
 Jacopo Nardiello <jacopo@sighup.io>
+Jason DeTiberus <detiber@gmail.com>
 Joe Beda <joe.github@bedafamily.com>
 Jonathan Frederickson <jonathan@terracrypt.net>
 Joshua Carp <jm.carp@gmail.com>
@@ -50,5 +51,6 @@ robjloranger <robjloranger@protonmail.com>
 S.Çağlar Onur <caglar@digitalocean.com>
 Steven Klassen <sklassen@gmail.com>
 Ted Wood <ted@xy0.org>
+Ulrich Schreiner <ulrich.schreiner@gmail.com>
 William Stewart <zoidbergwill@gmail.com>
 Xavier Lucas <xavier.lucas@corp.ovh.com>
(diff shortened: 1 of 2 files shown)


xmudrii:
Thanks for the contribution! Leaving `ssh-agent` changes to @kris-nova to test, as she had some problems in the past with it, so we want to make sure everything is working.

ulrichSchreiner:
As there is this comment in the code:
```go
// We don't need to check a key first, because SSH can support multiple auth implementations
// So here we just add BOTH a key based auth, and an arbitrary agent. Please don't touch this
// without talking to @kris-nova first OR unless something is just completely fucked
```
i first will ask before i try to modify the code again: i'm not really happy with the current code, because it **always** reads the private key and adds it as an `AuthMethod`  to the ssh configuration. But in my case (and i think in many people's case) the system agent would already contain the private key. So the entering of the passphrase (at least once after this PR) is not needed. 

i would try to change the code, so that the system uses the system agent and if it does not work, it uses the public key (without the `.pub`  extension) and adds this as an `AuthMethod`  to the ssh config.

to answer to the code comment: @kris-nova, no nothing is completely fucked, but it is really annoying to always enter the passphrase of my key although it is loaded in my agent :-). this PR solves the problem, that i have to enter the passphrase in every loop of the `RetryGetConfig` but it would be nicer to never enter it when it is not needed.

xmudrii:
> i first will ask before i try to modify the code again: i'm not really happy with the current code, because it always reads the private key and adds it as an AuthMethod to the ssh configuration. But in my case (and i think in many people's case) the system agent would already contain the private key. So the entering of the passphrase (at least once after this PR) is not needed.

I've this problem as well. As of the #542, I'm always getting asked for password to type in several times. However, the old code is not working for everybody, so we had to get rid of it.
This could look easy, but is actually a very complex topic due to various SSH keyring implementations.

We're trying to encourage contributors to contribute to the project, give ideas and open PRs, but when working on complex topics, it's recommended to talk to other contributors and maintainers.

If this code works for @kris-nova, I guess it's safe to merge it. The issue is very annoying, and this should fix it for most users.
I can ask Kris to take a look into that one on the developer call, and to talk about the problem, if we have some time. We already have a lot of stuff in the backlog, but I hope we'll be able to get everything sorted out.

kris-nova:
Typing in your password every time is completely fucked...

Thanks for the PR :heart:

PR #572, opened 2018-03-12
Api transition

kris-nova:
Gotta love a good transition.

Here is a working port to the cluster API.

1) Directives in the API aren't fleshed out (we use `providerConfig` for everything)
2) We aren't using CRD's yet (we just need to implement them as a state store)
3) We aren't running as a controller (but damn close)
4) We need to deprecate the autoscaling features (ASGs, etc)

Fixes #1 
Relates to #432 
Fixes #565 
Fixes #573
Fixes #471
Fixes #460

PR #574, opened 2018-03-12
get rid of the passphrase if key is already in agent

### LINKED ISSUE #577: User has to enter passphrase although key is loaded in ssh-agent
hi,

kubicorn uses `ssh` to connect to the master node of a cluster and query the kube-configuration. It uses the  key specified in the `SSH.publicKeyPath` from the configuration file. If the corresponding private key of this public key has a passphrase the user has to enter the phrase.

It is annoying for people using `ssh-agent` to enter this passphrase if the private key is already in the keyring of the agent, The only reason for running a agent is **not** to enter the passphrase, so kubicorn should not ignore the keys in the agent (it already connects to an agent if there is one running).

A better solution would be to first try to connect to a running agent and check if there is already a key provided by the agent which corresponds to the key specified in `publicKeyPath`.

ulrichSchreiner:
This PR simplifies the code and the behaviour of kubicorn. If the user has a running agent which has the needed key already in its ring, do not load the private key. So the user does not need to enter a passphrase. 

Closes #577

### DIFF
--- a/pkg/kubeconfig/kubeconfig.go
+++ b/pkg/kubeconfig/kubeconfig.go
@@ -43,25 +43,28 @@ type configLoader struct {
 
 func newConfigLoader(existing *cluster.Cluster) (*configLoader, error) {
 	pubKeyPath := local.Expand(existing.ProviderConfig().SSH.PublicKeyPath)
-	// --------------------------------------------------------------------------------
-	//
-	// @kris-nova
-	//
-	// We don't need to check a key first, because SSH can support multiple auth implementations
-	// So here we just add BOTH a key based auth, and an arbitrary agent. Please don't touch this
-	// without talking to @kris-nova first OR unless something is just completely fucked
-	//
+
+	// create an agent, either a connection to the one of the environment of the user or
+	// (if there is none) a go-implementation of a simple keyring
 	sshAgent := agent.NewAgent()
-	sshAgentWithKey, err := sshAgent.AddKey(pubKeyPath)
-	if err != nil {
-		return nil, fmt.Errorf("Unable to add key: %v", err)
-	}
 	sshConfig := &ssh.ClientConfig{
 		User:            existing.ProviderConfig().SSH.User,
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 	}
 	sshConfig.Auth = append(sshConfig.Auth, sshAgent.GetAgent())
-	sshConfig.Auth = append(sshConfig.Auth, sshAgentWithKey.GetAgent())
+
+	// check if pubkey is already in the agent. if it is already loaded
+	// in the agent the user does not need to enter the passphrase
+	if err := sshAgent.CheckKey(pubKeyPath); err != nil {
+		// add the key to the agent.
+		// THIS HAS A SIDEEFFECT: if the agent is the user's agent, the key will
+		// be appended to the list of keys. so the agent will have an additional
+		// key in it's ring. i don't see a big problem here.
+		_, err := sshAgent.AddKey(pubKeyPath)
+		if err != nil {
+			return nil, fmt.Errorf("Unable to add key: %v", err)
+		}
+	}
 	sshConfig.SetDefaults()
 
 	return &configLoader{


pkg/kubeconfig/kubeconfig.go
```diff
@@ -43,25 +43,28 @@ type configLoader struct {
...
 	sshConfig.Auth = append(sshConfig.Auth, sshAgent.GetAgent())
-	sshConfig.Auth = append(sshConfig.Auth, sshAgentWithKey.GetAgent())
+
+	// check if pubkey is already in the agent. if it is already loaded
+	// in the agent the user does not need to enter the passphrase
+	if err := sshAgent.CheckKey(pubKeyPath); err != nil {
+		// add the key to the agent.
+		// THIS HAS A SIDEEFFECT: if the agent is the user's agent, the key will
+		// be appended to the list of keys. so the agent will have an additional
+		// key in it's ring. i don't see a big problem here.
```
stealthybox: I don't see a problem here either. It's the same behavior as before.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -43,25 +43,28 @@ type configLoader struct {
...
-	sshAgentWithKey, err := sshAgent.AddKey(pubKeyPath)
-	if err != nil {
-		return nil, fmt.Errorf("Unable to add key: %v", err)
-	}
 	sshConfig := &ssh.ClientConfig{
 		User:            existing.SSH.User,
 		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
 	}
 	sshConfig.Auth = append(sshConfig.Auth, sshAgent.GetAgent())
-	sshConfig.Auth = append(sshConfig.Auth, sshAgentWithKey.GetAgent())
```
stealthybox: I think we need to add this second agent back at a minimum.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -43,25 +43,28 @@ type configLoader struct {
 
 func newConfigLoader(existing *cluster.Cluster) (*configLoader, error) {
 	pubKeyPath := local.Expand(existing.SSH.PublicKeyPath)
-	// --------------------------------------------------------------------------------
-	//
-	// @kris-nova
-	//
-	// We don't need to check a key first, because SSH can support multiple auth implementations
-	// So here we just add BOTH a key based auth, and an arbitrary agent. Please don't touch this
-	// without talking to @kris-nova first OR unless something is just completely fucked
```
stealthybox: We probably shouldn't get rid of this comment -- I remember @kris-nova went through some grief to add the second agent.

pkg/kubeconfig/kubeconfig.go
stealthybox: I misread this -- you're right, it's totally different.
The key isn't added to `sshAgentWithKey` anymore.

pkg/kubeconfig/kubeconfig.go
ulrichSchreiner: I'm not sure why this second line is needed. In the old code the key was added to the "sshAgent" too **and** an new keyring was returned which was stored as `sshAgentWithKey`. The code
```go
sshAgent.AddKey(pubKeyPath)
``` 
adds the key to `sshAgent` and also returns a new keyring which contains the key too.

So the loaded key was stored in `sshAgent` and in a second keyring --> the loaded key was inserted twice, in two agents. i do not see, why this should be necessary

stealthybox:
Supposedly, the second sshConfig entry allows users without an agent to SSH.
I don't see why this is necessary though since this function returns a new keyring if there is no SSH_AUTH_SOCKET or it fails to dial:
https://sourcegraph.com/github.com/kubicorn/kubicorn@0c682b572649cfee23a008c3b00270f931ae640d/-/blob/pkg/agent/agent.go#L136-141

Maybe there's a bug.

ulrichSchreiner:
`systemAgent` is called in `NewAgent` (which is the public interface). here the module creates a new *in memory* agent if the user has no system agent:
https://sourcegraph.com/github.com/kubicorn/kubicorn@0c682b572649cfee23a008c3b00270f931ae640d/-/blob/pkg/agent/agent.go#L51:6

kris-nova:
So in general the entire approach to this is an anti-pattern. We shouldn't even be using SSH here anyway and the SSH agent package we have isn't kind of a really weird implementation. I want to get rid of all this and do it the right way. 

Furthermore Kubicorn was completely broken for myself, and for others who didn't have SSH agent's set up. 

I am happy to test the code and if it makes sense I have no problems with merging it, but can we please follow protocol and have an issue open that explains what this is fixing?

ulrichSchreiner:
added issue #577

kris-nova:
If you want to rebase this we can merge it in - thanks for the PR

ulrichSchreiner:
rebased, but i think the failing tests are not my fault. the exist since #572 was merged

xmudrii:
‎Failing tests are okay, as they're not updated for #574 and the new cluster API. I'll be taking a look at this PR later today.

xmudrii:
This looks good to me. Let's give it a shoot and see how it works. If anybody has any problem, please open the issue, so we can research further.

@ulrichSchreiner Thanks for your contribution!

Issue #588, opened 2018-03-17
Packet: 422 You must specify a valid facility parameter


majst01:
Hi,

im following the packet instructions from here:
http://kubicorn.io/documentation/packet-walkthrough.html
I didnt modify the created _state/nameofproject/cluster.yaml 

But 
```
kubicorn apply -v 10 kubernetes-poc
```

ends with:
```
2018-03-17T15:21:44+01:00 [✖]  Error during apply of atomic reconciler, attempting clawback: POST https://api.packet.net/projects/<my_project_id>/devices: 422 You must specify a valid facility parameter
```
looking at the very beginning of the verbose log output already shows an empty Location:

```json
2018-03-17T15:21:43+01:00 [▶]  device.Apply devices actual 0, expected 1
2018-03-17T15:21:43+01:00 [▶]  creating server kubernetes-poc.master-0: packngo.DeviceCreateRequest{HostName:"kubernetes-poc.master-0", Plan:"baremetal_1", Facility:"", OS:"ubuntu_16_04", BillingCycle:"hourly", ProjectID:"fba6ca1f-0f69-4d7b-902d-1a9ad9dc10e4", UserData:"#!/usr/bin/env bash

2018-03-17T15:21:35+01:00 [ℹ]  Query existing resources
[...]

PR #589, opened 2018-03-17
Fix for #588, set location from ProviderConfig for a newResource to h…

majst01:
…ave the packet facility set which is a required parameter

### DIFF
--- a/cloud/packet/public/resources/device.go
+++ b/cloud/packet/public/resources/device.go
@@ -46,6 +46,14 @@ type Device struct {
 	ProjectID        string
 }
 
+func (r *Device) String() string {
+	return fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


cloud/packet/public/resources/device.go
```diff
@@ -46,6 +46,14 @@ type Device struct {
 	ProjectID        string
 }
 
+func (r *Device) String() string {
```
xmudrii: Where's this function being used and what exactly it does?

The PR looks good, thanks for the contribution! I would just like to clarify this before merging.

majst01: Without the Stringer debug log does not clearly show what value each property of Device has, i can remove if you want, but it makes logs a bit more readable here.

xmudrii: I think it's okay to stay for now.

xmudrii:
LGTM. Thank you for the contribution!

PR #591, opened 2018-03-18
Update kube-deploy to use ext-apiserver

### LINKED ISSUE #579: Rename the import to ext-apiserver
Right now we are vendoring an older version of `kube-deploy` [here](https://github.com/kubicorn/kubicorn/blob/master/apis/cluster/cluster.go#L22)

We need to use the new API [here](https://github.com/kubernetes/kube-deploy/blob/master/ext-apiserver/pkg/apis/cluster/v1alpha1/cluster_types.go)

tylerauerbeck:
Fixes #579


xmudrii:
Hi there! Thanks for the contribution!

To properly fix the issue, you need to vendor latest version of `kube-deploy`. We're using [`dep`](https://github.com/golang/dep) to manage dependencies, and [here's](https://github.com/kubicorn/kubicorn/blob/master/CONTRIBUTING.md#vendoring) our quick start guide for `dep`.

If you have any questions, let me know. :)

tylerauerbeck:
@xmudrii Still a little new to dep, so maybe you can point me in the right direction.

I've tried a few things (`dep ensure -add k8s.io/kube-deploy` and `dep ensure -add k8s.io/kube-deploy/ext-apiserver`) but it complained that it didn't have usable go code.

So instead I just did a `dep ensure -update`. This resulted in the following:

```
Solving failure: No versions of k8s.io/apimachinery met constraints:
....
release-1.8: package k8s.io/apimachinery/pkg/apis/meta/v1beta1 does not exist within project k8s.io/apimachinery
...
```

I'm assuming that fixing this would resolve the issue, but wasn't sure what the preferred method would be. v1beta1 doesn't exist within the apimachinery package until release-1.10. Up to that point, it's v1alpha1. Would the preference be to bump apimachinery to 1.10? Or to correct the code to use v1alpha1?

On the other hand, if there's a better way that I should be handling this, let me know.

xmudrii:
Sometimes, there can be some problems with `dep` and Kubernetes, but I was able to solve them in past [with something like this](https://github.com/kubicorn/kubicorn/pull/522/commits/a854425f4a85f21a84707a181603c75f00116acb#diff-836546cc53507f6b2d581088903b1785R70).

I was out of luck with using `releases-1.x` notation, however using revisions seemed like a solid workaround for me.

You can find to try some commit that contains the `v1alpha1` packages, and then to use that one with `dep`. Also, you can add entries manually to the `Gopkg.toml` file.

tylerauerbeck:
@xmudrii I can give that a try. But the question I still have is whether to move using v1beta1 or v1alpha1. V1beta1 seems to be what is currently coded, but would require updating Gopkg.toml. Reverting back to V1alpha1 would work fine give the current constraint on release-1.8, but I'm not sure if there is required functionality that would be present in V1beta1 that would be lost.

kris-nova:
I vote for getting it to vendor however we can, so long as it compiles ;) We can update later if we have a concrete reason to. If you need help with `dep` I am pretty good at dealing with dependencies

tylerauerbeck:
@kris-nova So looks like the repo changed from ext-apiserver to cluster-api today. So I updated that. Looks like one thing I'm struggling with right now is `dep ensure -update` is hanging on me and `dep ensure -add k8s.io/kube-deploy` complains about their being no usable go code.

tylerauerbeck:
@kris-nova @xmudrii So I at least got `dep` update to work, but it looks like this is causing its own set of problems when running `make compile`. My original issue seemed to be `bitbucket.org/ww/goautoneg` expecting mercurial to be present. Once I installed that package, everything started completing.  I'll try to find some time either tonight or tomorrow to _hopefully_ finish this up.

kris-nova:
Mind if I take a stab here? This is only a semantic change and the API definiton will be moving AGAIN in the coming weeks anyway. Thoughts?

tylerauerbeck:
@kris-nova Yepp no problem here. If you're able to pull this together, would love to know what I'm screwing up for future reference.

PR #598, opened 2018-03-21
[flags/cmd] place common flags into a set of functions that can be called to add them.

alejandroEsc:
without breaking pkg api contracts, allow for common flags to be pulled into function calls that add to a command those flags. Should remove a lot of duplication of code.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -63,18 +63,12 @@ func ApplyCmd() *cobra.Command {
 
 	fs := applyCmd.Flags()
 
-	fs.StringVarP(&ao.StateStore, keyStateStore, "s", viper.GetString(keyStateStore), descStateStore)
-	fs.StringVarP(&ao.StateStorePath, keyStateStorePath, "S", viper.GetString(keyStateStorePath), descStateStorePath)
-	fs.StringVarP(&ao.Set, keyKubicornSet, "e", viper.GetString(keyKubicornSet), descSet)
+	bindCommonStateStoreFlags(ao.StateStoreOptions, fs)
+	bindCommonAwsFlags(ao.AwsOptions, fs)
 
+	fs.StringVarP(&ao.Set, keyKubicornSet, "e", viper.GetString(keyKubicornSet), descSet)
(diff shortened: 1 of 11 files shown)


kris-nova:
Why are we doing this? Also what are we gaining? And why can't we just use `PersistentFlags`?

I agree that having the S3 flags everywhere is annoying, but I am wondering if having flags for this all together is the right approach. I would prefer to keep the flags in kubicorn to a minimum.

Thoughts?

alejandroEsc:
@kris-nova this is a follow up pr on a few things asked as part of the review process. In particular, here, it was asked if we could call flags to be added via functions instead of having to list the same ones over and over, therefore reduce chance of errors and simplify maintenance. In addition, you will notice that options have been organized in a cleaner manner, so we can establish a nice pattern to follow.

As to your question about having flags being the right approach, im not sure what you're thinking? Personally i would rather have the option of a config file to be used and perhaps the use environment variables. If this is something you think we should do, i think @adnxn is working on adding a config file? As for environment variables, those are there, we can then just remove flags in general. hmmm or perhaps you mean move the flags a layer above individual commands?

adnxn:
> Personally i would rather have the option of a config file to be used and perhaps the use environment variables. If this is something you think we should do, i think @adnxn is working on adding a config file? 

yes i am, and i think my config changes will cut down on the number of flags we're using here. 

> As for environment variables, those are there, we can then just remove flags in general. 

so yea, I think we're going to move towards just environment variables and config file.

alejandroEsc:
@adnxn @kris-nova feel free to close if we are not going this route. Thanks!

xmudrii:
@alejandroEsc I'll check with @kris-nova once again. I think we can get it in, and then decide where to go next.

If you want, you can rebase the PR.

xmudrii:
@alejandroEsc Thanks for rebasing the PR! Can you please gofmt code as well?

alejandroEsc:
@xmudrii thanks for doing the check, yikes i should have done that. Ok will look into your suggestions!

xmudrii:
Relevant issue #617

PR #601, opened 2018-03-22
Fixing version command

### LINKED ISSUE #593: Improve GetVersion() logic
Once we start publishing releases, current [`GetVersion()`](https://github.com/kubicorn/kubicorn/blob/master/cmd/version.go#L67) is not going to work, because this assumes users have installed and configured Go environment, and they have `kubicorn` cloned.

A potential solution to this is to compile-in version using compiler flags when we build binaries.

kris-nova:
closes #593


nrb:
Do we need to set the `LDFLAGS` vars? Here's how Ark does it in their build script.

https://github.com/heptio/ark/blob/master/hack/build.sh#L52

kris-nova:
Instead of defining LDFLAGS I just munged all of the `go build` stuff into a single var in https://github.com/kubicorn/kubicorn/pull/601/commits/b9068155cf1d2a47e159b1dda9d0dd1c5c197dce

Should fix the problem 😄

kris-nova:
Do we have a preference for LDFLAGS - I have no strong opinions here

nrb:
> Do we have a preference for LDFLAGS - I have no strong opinions here

I don't.

PR #615, opened 2018-04-01
Configure git to use the https protocol instead of ssh

Siilwyn:
I couldn't get the docker build to work locally because I don't have access to the kubicorn repository over SSH. Changing it to use https instead fixes this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@
 
 FROM golang:latest
 
+RUN git config --global url."https://${GITHUB_OAUTH_TOKEN}@github.com/".insteadOf "git@github.com:"
+
 # Create the default data directory
 RUN go get github.com/kubicorn/kubicorn/...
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,6 +5,8 @@
 
 FROM golang:latest
 
+RUN git config --global url."https://${GITHUB_OAUTH_TOKEN}@github.com/".insteadOf "git@github.com:"
```
xmudrii: Before I merge the PR, I would like to know little bit more about the `GITHUB_OAUTH_TOKEN` variable. Is this something specific to CircleCI, or something from GitHub app? Can we expect it to be set for everybody?

This works with CircleCI, but I want to make sure the Dockerfile will work for all users.

Siilwyn: Whoops, sharp one! That's a leftover I will remove, copied this from a Go dockerfile where this token is needed to access private repositories.

xmudrii:
@Siilwyn Looks good to me. Thanks for the contribution!

Issue #533, opened 2018-02-04
teleconsole and telehost


gedw99:
do you think that putting in telehost as a bastion in the cluster makes sense ?

I like it allot

PR #619, opened 2018-04-08
ssh,scp: implement ssh and scp clients

### LINKED ISSUE #609: `kubicorn ssh` cli command
Kubicorn is looking nice nowadays =D I love that you echo out the command to shell into the nodes on creation, however if I don't write that command down its hard to figure out how to do it again.

I propose a `kubicorn ssh` command that just shell execs the needed command for you to make it easy to connect into the nodes.

xmudrii:
# `ssh` and `scp` packages

Closes #609

This PR implements the `ssh` package used to handle SSH connections. It comes with a `kubicorn ssh <cluster-name>` CLI command for connecting to the master instance of the cluster.

Beside `ssh` package and command, this PR refactors the `scp` package to utilize the new `ssh` package, as well as implements the `WriteBytes` function used to send a file to the cluster.

## Implementation of the PR

Currently, `ssh` package is used only for `kubicorn ssh` command.
The `scp` package is used to download the VPN credentials, if VPN is enabled for DigitalOcean clusters.

## Resolving `agent` problems

This `ssh` implementation is not dependent on the [`agent`](https://github.com/kubicorn/kubicorn/tree/master/pkg/agent) package we're using currently. Instead, it implements the `auth` package which handles SSH authentication.

The default authentication method is password authentication. As we're connecting only one time (ensured by the `Connect` function) and then we're just reusing the same connection, user needs to enter password only once.

Beside password authentication, users can also use the system SSH agent. To use the system agent, it's required to set the `KUBICORN_SSH_AGENT` environment variable (documented in `envar.md`). If the variable is set, `kubicorn` will try to use system SSH agent. If error occurs (e.g. the system agent is not present or the key is not present in the agent), `kubicorn` will resort to the password authentication with the appropriate error message and recommended steps to fix the issue.

## Notes

* The `kubicorn ssh` command supports only SSH-ing to the master.

## Parts needing attention

* [`StartInteractiveSession` - `xterm` TTY](https://github.com/kubicorn/kubicorn/pull/619/files#diff-719664d98cb82359b9997ba63b0e7ffdR146). What exactly the `RequestPty` function does and is it safe to use `xterm` for all clusters?
* [`StartInteractiveSession` - handle terminal window changes](https://github.com/kubicorn/kubicorn/pull/619/files#diff-719664d98cb82359b9997ba63b0e7ffdR151). Without that part, if user resizes the terminal, SSH session will misbehave. I'm not sure is it okay to run this in an infinitive `for` loop in the goroutine. Do we need a timeout for goroutine here?

## Steps before merging PR

We want to make this actually works. I would love @kris-nova to test this before we merge the PR, so we're sure it works as intended. The easiest way to test the PR is to use the `kubicorn ssh` command.

## Steps after merging PR

To make this PR cleaner and easier to review, it only implements the `ssh` and `scp` packages without implementing them in the code (beside for DigitalOcean VPN). When the PR is merged, we need to implement it for the [`kubeconfig`](https://github.com/kubicorn/kubicorn/blob/master/pkg/kubeconfig/kubeconfig.go) and [`initapi`](https://github.com/kubicorn/kubicorn/blob/master/pkg/initapi/ssh.go) packages.

After merging the PR, we can make the following changes to further improve the packages:

* Make the `kubicorn ssh` command able to SSH to all nodes in the cluster.
* Implement agent forwarding (probably hidden behind an environment variable). Helpful if you want to be able to use your agent/SSH keys from the cluster.
* Delete the `agent` package when we implement the `ssh` package in all places.
* Possibly integrate with `teleport` and `teleconsole` packages and offer profiles with bastion (issue #533).


cmd/ssh.go
```diff
@@ -0,0 +1,120 @@
...
+		ip, err = ipresolver.GetNodeIPAddress(options.NodeName)
+		if err != nil {
+			return fmt.Errorf("Unable to obtain IP address of node [%s] in cluster [%s]: %v",
+				options.NodeName, options.Name, err)
+		}
+	} else {
+		ip = cluster.ProviderConfig().KubernetesAPI.Endpoint
+	}
+
+	client := ssh.NewSSHClient(ip, "22", "root")
```
kris-nova: Should we hard code this? I know the user and port is stored in the cluster API...

xmudrii: It should be better now :)

xmudrii:
@kris-nova ready for review

pkg/ssh/auth/agent.go
```diff
@@ -0,0 +1,61 @@
...
+	"io/ioutil"
+	"net"
+	"os"
+
+	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
+)
+
+// SystemAgent returns system agent if it exists.
+func SystemAgent() agent.Agent {
```
kris-nova: So wondering how the workflow would work here..

Should we return an error if the system agent doesnt exist? 
Should we change the function name to `SystemAgentIfExists()` or something?

Just think we could be a little more strict here is all

cloud/digitalocean/droplet/resources/droplet.go
```diff
@@ -171,17 +168,21 @@ func (r *Droplet) Apply(actual, expected cloud.Resource, immutable *cluster.Clus
...
-				scp := scp.NewSecureCopier(immutable.ProviderConfig().SSH.User, masterIPPublic, "22", privPath, agent)
-				masterVpnIP, err := scp.ReadBytes("/tmp/.ip")
+				client := ssh.NewClient(masterIPPublic, providerConfig.SSH.Port,
+					providerConfig.SSH.User, providerConfig.SSH.PublicKeyPath)
+				err = client.Connect()
+				if err != nil {
+					return nil, nil, fmt.Errorf("Unable to connect to SSH: %v", err)
+				}
+
+				masterVpnIP, err := scp.ReadBytes(client, "/tmp/.ip")
```
kris-nova: Do we want to pass in the client every time we call `ReadBytes` or do we want to cache it in memory somewhere and then tell it which file to read from?

pkg/ssh/auth/agent.go
xmudrii: > Should we return an error if the system agent doesnt exist?

It could be useful for debugging purposes, so I added it.

> Should we change the function name to SystemAgentIfExists() or something?

Sounds better than `SystemAgent`. Updated.

cloud/digitalocean/droplet/resources/droplet.go
xmudrii: I guess we can encapsulate it into a struct. I've updated the SCP package to try to fix this. Let me know is it any better.

PR #631, opened 2018-04-17
WIP: AWS private topology

mgoodness:
Creates a new API object named `PublicSubnet`, and adds a collection of them to the `Network` object. NAT Gateways are created in these, which are used as the default gateways (instead of the Internet Gateway) for the node subnets.

The current logic requires one PublicSubnet in the same AZ as a node subnet. For example, a node subnet in us-west-2a will be configured to use a NAT Gateway created in a us-west-2 PublicSubnet. RouteTable creation will fail if this requirement is not met.

Also added some `WaitUntil*` calls on resource creation that could be helpful in the public AWS provider.

TODO (not exhaustive):
- [ ] Create an ELB and a Route53 record so the API server is accessible from outside the VPC
- [x] Rename PublicSubnet and allow them to be public or private (for use by external/internal ELBs)


mgoodness:
There are now `PublicSubnets` and `PrivateSubnets` in the `Network` Object. As before, NAT Gateways are created in the `PublicSubnets`. They're also tagged with `kubernetes.io/role/elb = true`. `PrivateSubnets` are tagged with `kubernetes.io/role/internal-elb = true`.

xmudrii:
@mgoodness Thank you so much for the contribution! We think this looks great, and that it could be merged.

I'm not sure what's status of the PR, but looks like it needs rebase to the `Gopkg.lock` file. Not sure, but I think it could be fixed with `dep ensure`.

If you want to fix this, I think we could get this PR in. Than, if you would like, you can create a follow-up PR with the improvements.

Thanks once again!

Issue #651, opened 2018-07-11
intermittent failures


pbarker:
Hey folks, we are seeing intermittent failures running kubicorn in a pipeline.

```
2018-07-11T12:38:41Z [✖]  Error during apply of atomic reconciler, attempting clawback: Unable to tag new Subnet: InvalidSubnetID.NotFound: The subnet ID 'subnet-09fb22c5c3e61b5d3' does not exist
	status code: 400, request id: b7a98721-f314-454f-ba68-82afdb863b1f
```
do we need a retry loop there?

mattmoyer:
Here are a couple more similar intermittent failures we've seen (I work w/@pbarker):

- Internet gateway does not exist yet (need to retry on the attachment):
   ```
   2018-07-12T12:58:03Z [ℹ]  Attaching Internet Gateway [igw-0251558381aa8669c] to VPC [vpc-04c3a5926a120f250]
   2018-07-12T12:58:03Z [✖]  Error during apply of atomic reconciler, attempting clawback: InvalidInternetGatewayID.NotFound: The internetGateway ID 'igw-0251558381aa8669c' does not exist
	status code: 400, request id: fc2a3fc2-0e43-405e-b01f-e504fb7ce0e5
   ```

- Route table does not exist (need to retry route table mapping):
   ```
   2018-07-12T13:10:40Z [ℹ]  Mapping route table [rtb-04e09786378cda577] to internet gateway [igw-0969c5f82a3321216]
   2018-07-12T13:10:40Z [✖]  Error during apply of atomic reconciler, attempting clawback: InvalidRouteTableID.NotFound: The routeTable ID 'rtb-04e09786378cda577' does not exist
	status code: 400, request id: 5defb538-e357-49f2-89eb-00b4783cd9aa
```

pbarker:
thanks @mattmoyer, the common thread here is that they are all 400's. I mention https://github.com/kubicorn/kubicorn/pull/652#issuecomment-404248970 that the default retryer doesn't retry on 400's, this looks like sufficient reason to implement a custom retryer.

kris-nova:
@pbarker @mattmoyer feel free to ping me if you need help getting this through - im not combing the issue tracker like I once was

PR #652, opened 2018-07-11
adds explicit retries to subnet tagging

pbarker:
fixes https://github.com/kubicorn/kubicorn/issues/651

### DIFF
--- a/cloud/amazon/public/resources/subnet.go
+++ b/cloud/amazon/public/resources/subnet.go
@@ -16,6 +16,7 @@ package resources
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/kubicorn/kubicorn/apis/cluster"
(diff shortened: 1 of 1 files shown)


xmudrii:
@pbarker If you can get `pkg/retry`, I think that's okay. Maybe you can also wrap it around some `for` loop, and just retry like that. You can find such pattern for instances I think.

cloud/amazon/public/resources/subnet.go
```diff
@@ -247,7 +248,12 @@ func (r *Subnet) tag(tags map[string]string) error {
 			Value: S("%s", val),
 		})
 	}
-	_, err := Sdk.Ec2.CreateTags(tagInput)
+	req, _ := Sdk.Ec2.CreateTagsRequest(tagInput)
+	retry := true
+	req.Retryable = &retry
+	req.RetryCount = 30
```
kris-nova: Ideally these would go up as constants

kris-nova: If you want to update feel free - otherwise I will just merge - LMK what you think

kris-nova:
It's a shame we live in a world where SDKs need to come with built in retry loops - but here we are. 

I'd rather use theirs, than ours. Especially because the AWS API can be tricky.

LGTM

PR #658, opened 2018-07-27
Support multi cluster kubeconfig

ashish-amarnath:
Signed-off-by: Ashish Amarnath <ashish.amarnath@gmail.com>

What this PR does / why we need it:
This PR implements support for multicluster kubeconfig fixing https://github.com/kubicorn/kubicorn/issues/496

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes # 496

### DIFF
--- a/pkg/kubeconfig/kubeconfig.go
+++ b/pkg/kubeconfig/kubeconfig.go
@@ -28,6 +28,9 @@ import (
 	"github.com/kubicorn/kubicorn/pkg/logger"
 	"github.com/pkg/sftp"
 	"golang.org/x/crypto/ssh"
+
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
 const (
@@ -72,6 +75,33 @@ func newConfigLoader(existing *cluster.Cluster) (*configLoader, error) {
 		sshconfig: sshConfig,
 	}, nil
 }
+
+func getRemoteFileBytes(address, filePath string, sshCfg *ssh.ClientConfig) ([]byte, error) {
+	conn, err := ssh.Dial("tcp", address, sshCfg)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+	c, err := sftp.NewClient(conn)
+	if err != nil {
+		return nil, err
+	}
+	defer c.Close()
+	r, err := c.Open(filePath)
+	if err != nil {
+		return nil, err
+	}
+	defer r.Close()
+	return ioutil.ReadAll(r)
+}
+
+func getRemoteKubeconfigPath(user string) string {
+	if user == "root" {
+		return "/root/.kube/config"
+	}
+	return filepath.Join("/home", user, ".kube/config")
+}
+
 func (cfg *configLoader) GetConfig() error {
 	user := cfg.existing.ProviderConfig().SSH.User
 	if cfg.existing.ProviderConfig().SSH.Port == "" {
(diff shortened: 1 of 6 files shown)


pkg/kubeconfig/kubeconfig.go
```diff
@@ -72,6 +75,41 @@ func newConfigLoader(existing *cluster.Cluster) (*configLoader, error) {
...
+	defer r.Close()
+	bytes, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, err
+	}
+
+	return bytes, nil
+}
+
+func getUserKubeconfigPath(user string) string {
```
xmudrii: I would rename this function to prevent confusion. Initially, I thought `getUserKubeconfigPath` is path to kubeconfig on the local (user's) machine.

I think something along `remoteKubeconfigPath` would sound better.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -72,6 +75,41 @@ func newConfigLoader(existing *cluster.Cluster) (*configLoader, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	defer c.Close()
+	r, err := c.Open(filePath)
+	if err != nil {
+		return nil, err
+	}
+	defer r.Close()
+	bytes, err := ioutil.ReadAll(r)
```
xmudrii: Maybe we could simplify this with `return ioutil.ReadAll(r)`.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -126,17 +148,22 @@ func (cfg *configLoader) GetConfig() error {
...
 		}
+	} else {
+		existingKubeConfig, err = clientcmd.LoadFromFile(localPath)
+		if err != nil {
+			return err
+		}
 	}
 
-	f, err := os.OpenFile(localPath, os.O_APPEND|os.O_WRONLY, os.ModeAppend)
+	mergedKubeConfigBytes, err := clientcmd.Write(*mergeKubeconfigs([]*clientcmdapi.Config{existingKubeConfig, newKubeConfig}))
```
xmudrii: We could simplify this with [`clientcmd.WriteToFile`](https://godoc.org/k8s.io/client-go/tools/clientcmd#WriteToFile).

pkg/kubeconfig/kubeconfig.go
ashish-amarnath: Agree.

pkg/kubeconfig/kubeconfig.go
ashish-amarnath: Absolutely, that would be a good simplification.

pkg/kubeconfig/kubeconfig.go
ashish-amarnath: Good catch. Thanks :)

pkg/kubeconfig/kubeconfig.go
```diff
@@ -72,6 +75,36 @@ func newConfigLoader(existing *cluster.Cluster) (*configLoader, error) {
...
+	defer c.Close()
+	r, err := c.Open(filePath)
+	if err != nil {
+		return nil, err
+	}
+	defer r.Close()
+	return ioutil.ReadAll(r)
+}
+
+func getRemoteKubeconfigPath(user string) string {
```
xmudrii: This function could be simplified little bit, such as:
```
if user == "root" {
    return "/root/.kube/config"
}
return filepath.Join("/home", user, ".kube/config")
```
There is no need for another variable and more complex if/else logic.

pkg/kubeconfig/kubeconfig.go
```diff
@@ -126,19 +143,14 @@ func (cfg *configLoader) GetConfig() error {
...
-		return err
-	}
-	_, err = f.WriteString(string(bytes))
-	if err != nil {
-		return err
-	}
-	defer f.Close()
-	logger.Always("Wrote kubeconfig to [%s]", localPath)
-	return nil
+	return clientcmd.WriteToFile(*mergeKubeconfigs([]*clientcmdapi.Config{existingKubeConfig, newKubeConfig}), localPath)
```
xmudrii: I'm wondering could we improve this. If the `WriteToFile` function takes struct as it is, and not a pointer to it. maybe we can just make `mergeKubeconfigs` return struct instead of a pointer.

Potentially the struct is not that big, and we're not modifying it all, so that could work and be okay performance wise.

Not a big thing, it can stay both ways, but looking can we make this more readable. :)

ashish-amarnath: I wanted to keep the return type of `mergeKubeconfigs` consistent with that of `clientcmdapi.NewConfig()`. This is also consistent with most `clientcmd` loader functions.
The pattern I've seen here is, functions return a pointer but take parameters by-value. For that reason, if it is OK with you, I'll keep this as is.

pkg/kubeconfig/kubeconfig.go
ashish-amarnath: yes, thanks for the simplification. This looks so much better now. :)

pkg/kubeconfig/kubeconfig.go
ashish-amarnath: Although, I agree with you on the readability of this code and I have made it more readable. LMK WYT.

xmudrii: Right, that makes a lot of sense. I like how it looks right now, so let's leave it as it is.

PR #671, opened 2018-08-06
Add kubeconfig-local-file flag to create

ncdc:
Signed-off-by: Andy Goldstein <andy.goldstein@gmail.com>


cmd/create.go
```diff
@@ -19,6 +19,8 @@ import (
 	"os"
 	"strings"
 
+	"github.com/kubicorn/kubicorn/pkg/kubeconfig"
```
xmudrii: nit: import order

ncdc: Yay `goimports`

PR #677, opened 2018-09-12
Added a `CloudManagerSecret` to Cluster API to enable CloudManagers

Richard87:
Added a `CloudManagerSecret` to Cluster API to enable CloudManagers, and specifically add DigitalOceans Cloud-Manager and CSI

Related PR for Bootstrap scripts: https://github.com/kubicorn/bootstrap/pull/23

### DIFF
--- a/apis/cluster/cluster.go
+++ b/apis/cluster/cluster.go
@@ -16,10 +16,11 @@ package cluster
 
 import (
 	"encoding/json"
-	"k8s.io/api/core/v1"
 
 	"github.com/kubicorn/kubicorn/pkg/logger"
+
 	appsv1beta2 "k8s.io/api/apps/v1beta2"
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "k8s.io/kube-deploy/cluster-api/api/cluster/v1alpha1"
 )
@@ -64,7 +65,7 @@ type Cluster struct {
 	// The pattern here says that from an arbitrary deployment, you should be able
 	// to bootstrap anything else you could need. We default to the kubicorn controller.
 	ControllerDeployment *appsv1beta2.Deployment `json:"controllerDeployment,omitempty"`
-	CloudManagerSecret   *v1.Secret              `json:"cloudManagerSecret,omitempty"`
+	APITokenSecret       *v1.Secret              `json:"cloudManagerSecret,omitempty"`
 }
 
 // ProviderConfig is a convenience method that will attempt
--- a/apis/cluster/cluster.go
+++ b/apis/cluster/cluster.go
@@ -16,6 +16,7 @@ package cluster
 
 import (
 	"encoding/json"
+	"k8s.io/api/core/v1"
 
 	"github.com/kubicorn/kubicorn/pkg/logger"
 	appsv1beta2 "k8s.io/api/apps/v1beta2"
@@ -63,6 +64,7 @@ type Cluster struct {
 	// The pattern here says that from an arbitrary deployment, you should be able
 	// to bootstrap anything else you could need. We default to the kubicorn controller.
 	ControllerDeployment *appsv1beta2.Deployment `json:"controllerDeployment,omitempty"`
+	CloudManagerSecret   *v1.Secret              `json:"cloudManagerSecret,omitempty"`
 }
 
 // ProviderConfig is a convenience method that will attempt
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -208,6 +208,24 @@ func runApply(options *cli.ApplyOptions) error {
 		crdStateStore.Commit(newCluster)
 	}
 
+	if newCluster.CloudManagerSecret != nil {
+		// -------------------------------------------------------------------------------------------------------------
+		//
+		// Here is where we hook in for the new controller logic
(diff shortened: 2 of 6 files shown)


Richard87:
This is not ready, I realised the Secret gets "double" base64 encoded, so I need to tweak it a little bit :)

apis/cluster/cluster.go
```diff
@@ -16,6 +16,7 @@ package cluster
 
 import (
 	"encoding/json"
+	"k8s.io/api/core/v1"
```
xmudrii: This should be moved to the second (below) group.

apis/cluster/cluster.go
```diff
@@ -63,6 +64,7 @@ type Cluster struct {
 	// The pattern here says that from an arbitrary deployment, you should be able
 	// to bootstrap anything else you could need. We default to the kubicorn controller.
 	ControllerDeployment *appsv1beta2.Deployment `json:"controllerDeployment,omitempty"`
+	CloudManagerSecret   *v1.Secret              `json:"cloudManagerSecret,omitempty"`
```
xmudrii: We could remove this field to something like `APITokenSecret`, so it's easier to understand what's stored here.

cmd/apply.go
```diff
@@ -208,6 +208,24 @@ func runApply(options *cli.ApplyOptions) error {
...
+		//
+		// Here is where we hook in for the new controller logic
+		// This is exclusive to profiles that have a cloud-manager secret defined
+		//
+		logger.Info("Deploying cloud manager secret: %s", newCluster.CloudManagerSecret.ObjectMeta.Name)
+		err = resourcedeploy.DeployCloudManagerSecret(newCluster)
+		if err != nil {
+			return fmt.Errorf("Unable to deploy cloud manager secret: %v", err)
+		}
+		crdStateStore := crd.NewCRDStore(&crd.CRDStoreOptions{
```
xmudrii: This is not needed, along with `crdStateStore.Commit(newCluster)`. This is used by the `cluster-api` and is not needed in this case.

profiles/digitalocean/centos_7.go
```diff
@@ -142,5 +145,15 @@ func NewCentosCluster(name string) *cluster.Cluster {
 	c := cluster.NewCluster(name)
 	c.SetProviderConfig(controlPlaneProviderConfig)
 	c.NewMachineSetsFromProviderConfigs(machineSetsProviderConfigs)
+
+	secret := &apiv1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "digitalocean",
+			Namespace: "kube-system",
+		},
+		Data: map[string][]byte{"access-token": []byte(os.Getenv("DIGITALOCEAN_ACCESS_TOKEN"))},
```
xmudrii: I think you can solve the double-base64 encoding by using `StringData` instead of `Data` here.

Richard87:
Fixed the comments, ad tested a little more

`do` , `digitalocean` and  `do-ubuntu` works nicely :)
`do-centos` failed without any events in `/var/log/cloud-init-output.log` (the file was missing).
`cdo` or `controller-digitalocean-ubuntu` is not tested


It's tested by running this yaml:
```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: csi-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
  storageClassName: do-block-storage
---
kind: Pod
apiVersion: v1
metadata:
  name: my-csi-app
spec:
  containers:
  - name: my-frontend
    image: busybox
    volumeMounts:
    - mountPath: "/data"
      name: my-do-volume
    command: [ "sleep", "1000000" ]
  volumes:
  - name: my-do-volume
    persistentVolumeClaim:
      claimName: csi-pvc
```
The purpose is to check the CSI and volumes work, and that CSI-DO can read the secret.

Richard87:
New smoketest to also test that DigitalOcean Cloud Manager creates the required LoadBalancer's (works nicely!)

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: csi-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
  storageClassName: do-block-storage
---
kind: Pod
apiVersion: v1
metadata:
  name: my-csi-app
spec:
  containers:
  - name: my-frontend
    image: busybox
    volumeMounts:
    - mountPath: "/data"
      name: my-do-volume
    command: [ "sleep", "1000000" ]
  volumes:
  - name: my-do-volume
    persistentVolumeClaim:
      claimName: csi-pvc
---
apiVersion: v1
kind: Service
metadata:
  name: httpbin
  labels:
    app: httpbin
spec:
  type: LoadBalancer
  ports:
  - name: http
    port: 8000
  selector:
    app: httpbin
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin
spec:
  replicas: 3
  selector:
    matchLabels:
      app: httpbin
  template:
    metadata:
      labels:
        app: httpbin
        version: v1
    spec:
      containers:
      - image: docker.io/citizenstig/httpbin
        imagePullPolicy: IfNotPresent
        name: httpbin
        ports:
        - containerPort: 8000
```

Richard87:
The loadbalancer and deployments works good. 

*BUT* the Firewall that is created blocks all traffic, but that might be a problem on DO's end

PR #681, opened 2018-10-13
Fix golint build issue (wrong import)

PaulMaddox:
This PR fixes the golint import path, which is currently causing travis builds to fail.


kris-nova:
Thanks! Mountain pic per our agreement is on Twitter - hard to upload via mobile!

PR #685, opened 2018-11-08
AWS: allow security group members to communicate

ncdc:
For multi-node configurations, all members of the same security group
should be able to communicate with each other. Without this change,
communication between control plane nodes and worker nodes is
functional, but communication from one control plane node to another
control plane node is blocked, as is worker node to worker node. Overlay
networks such as Calico require node-to-node communications, and this
change allows these networks to function.

Signed-off-by: Andy Goldstein <andy.goldstein@gmail.com>

### DIFF
--- a/cloud/amazon/public/resources/securitygroup.go
+++ b/cloud/amazon/public/resources/securitygroup.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


ncdc:
This may be too broad. Would appreciate eyes from @detiber if he's got some time.

PR #690, opened 2019-05-03
upgrade golang to 1.9.3 to solve compile issue

figo:
otherwise, an error pop up:
vendor/github.com/json-iterator/go/config.go:107: undefined: sync.Map

Signed-off-by: Hui Luo <luoh@vmware.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ endif
 PKGS=$(shell go list ./... | grep -v /vendor)
 CI_PKGS=$(shell go list ./... | grep -v /vendor | grep -v test)
 FMT_PKGS=$(shell go list -f {{.Dir}} ./... | grep -v vendor | grep -v test | tail -n +2)
-SHELL_IMAGE=golang:1.8.3
+GOLANG_IMAGE=golang:1.9.3
 GIT_SHA=$(shell git rev-parse --verify HEAD)
 VERSION=$(shell cat VERSION)
 PWD=$(shell pwd)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ endif
 PKGS=$(shell go list ./... | grep -v /vendor)
 CI_PKGS=$(shell go list ./... | grep -v /vendor | grep -v test)
 FMT_PKGS=$(shell go list -f {{.Dir}} ./... | grep -v vendor | grep -v test | tail -n +2)
-SHELL_IMAGE=golang:1.8.3
+SHELL_IMAGE=golang:1.9.3
```
detiber: Would this build with a newer version of golang or is`1.9.3` the latest that will work without changes?

Makefile
```diff
@@ -52,7 +52,7 @@ build-linux-amd64: ## Create the kubicorn executable for Linux 64-bit OS in the
 	-w /go/src/github.com/kubicorn/kubicorn \
 	-v ${PWD}:/go/src/github.com/kubicorn/kubicorn \
 	-e GOPATH=/go \
-	--rm golang:1.8.1 make docker-build-linux-amd64
+	--rm golang:1.9.3 make docker-build-linux-amd64
```
detiber: Should this possible reference the SHELL_IMAGE variable instead of hardcoding it here?

Makefile
figo: i did not test further, 1.9.* is probably the minimum upgrade to pass compiling, the e2e test based on kubicorn binary (not in this repo) passed after this change.

also, there is a discussion around this seems to confirm the theory:
https://github.com/kubernetes/test-infra/issues/7183

Makefile
figo: sounds good, i'll change it

Issue #686, opened 2019-02-27
Submodule is using ssh rather than https


joedborg:
Just wondering if this is done intentially?  I’m trying to build a snap package and I don’t want to have to generate ssh keys for that build node.

dminca:
Confirming this one, I've just did a `go get github.com/kubicorn/kubicorn` and it failed:

```sh
# cd ~/Repos/open-source/others/gopath/src/github.com/kubicorn/kubicorn; git submodule update --init --recursive
Submodule 'bootstrap' (git@github.com:kubicorn/bootstrap.git) registered for path 'bootstrap'
Cloning into '~/Repos/open-source/others/gopath/src/github.com/kubicorn/kubicorn/bootstrap'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:kubicorn/bootstrap.git' into submodule path '~/Repos/open-source/others/gopath/src/github.com/kubicorn/kubicorn/bootstrap' failed
Failed to clone 'bootstrap'. Retry scheduled
Cloning into '~/Repos/open-source/others/gopath/src/github.com/kubicorn/kubicorn/bootstrap'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:kubicorn/bootstrap.git' into submodule path '~/Repos/open-source/others/gopath/src/github.com/kubicorn/kubicorn/bootstrap' failed
Failed to clone 'bootstrap' a second time, aborting
package github.com/kubicorn/kubicorn: exit status 1
```

PR #693, opened 2019-08-02
fix issue with submodule protocol - cloning over SSH instead of HTTPS

dminca:
When executing the install command from the docs it's going to fail
because it's trying to also fetch the submodule, and it was doing that
via SSH instead of HTTPS protocol.

As a result, replace the submodule's protocol with HTTPS so that anyone
can freely issue `go get github.com/kubicorn/kubicorn` and have the binary directly compiled in their `${GOPATH}`

Resolves: #686
Related:
Signed-off-by: Daniel Andrei Minca

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "bootstrap"]
 	path = bootstrap
-	url = https://github.com:kubicorn/bootstrap.git
+	url = https://github.com/kubicorn/bootstrap.git
+
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "bootstrap"]
 	path = bootstrap
-	url = git@github.com:kubicorn/bootstrap.git
+	url = https://github.com:kubicorn/bootstrap.git


detiber:
I don't have issues with this change, but it does potentially complicate developer workflow.

@xmudrii thoughts?

xmudrii:
@detiber Not sure do gitmodules need write rights, but if not, it should improve the workflow as git over http for public repos should always succeed (when reading).

If it needs to write anything to remote, then it's going to complicate quite a bit, but as far as I know it should not do such thing.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "bootstrap"]
 	path = bootstrap
-	url = git@github.com:kubicorn/bootstrap.git
+	url = https://github.com:kubicorn/bootstrap.git
```
xmudrii: Shouldn't this be `https://github.com/...` (`/` instead of `:`)?

dminca: well spotted 👍

dminca:
You're welcome 👍

